PR #9, opened 2023-03-22
Enable macos arm64 builds

Shelnutt2:
Enable macos arm64 builds

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -1,1 +1,13 @@
-{}
+# - only affects README installation instructions
+# - edit recipe/conda_build_config.yaml to affect build
+channels:
+  targets:
+    - ["tiledb", "main"]
+github:
+  user_or_org: TileDB-Inc
+  branch_name: master
+  tooling_branch_name: main
+build_platform:
+  osx_arm64: osx_64
+conda_forge_output_validation: true
+test_on_native_only: true
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,29 +1,36 @@
 {% set name = "tiledbsoma" %}
 {% set version = "1.1.1" %}
+{% set sha256 = "ac01357ed1f793fe96f728e4abb4965c7303918510cec38fda1ea5ef5ad47643" %}
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 source:
-    git_url: https://github.com/single-cell-data/TileDB-SOMA.git
-    git_rev: b0b91f6d875f71a542859d944aae1fc82c107862
+  #  git_url: https://github.com/single-cell-data/TileDB-SOMA.git
+  #  git_rev: b0b91f6d875f71a542859d944aae1fc82c107862
+  url: https://github.com/TileDB-Inc/single-cell-data/TileDB-SOMA/{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 1
+  skip: true  # [linux32 or py2k]
 
 outputs:
   - name: libtiledbsoma
     version: {{ version }}
-    script: build-libtiledbsoma.sh
+    script: build-libtiledbsoma.sh  # [not win]
     requirements:
       build:
         - {{ compiler('cxx') }}
         - cmake
-        - make
+        - make  # [not win]
         - pybind11
       host:
-        - tiledb 2.14.*
+        - tiledb 2.15.*
         - spdlog
       run:
-        - tiledb 2.14.*
+        - tiledb 2.15.*
         - spdlog
 
   - name: tiledbsoma-py
@@ -36,6 +43,11 @@ outputs:
     requirements:
       build:
         - {{ compiler('cxx') }}
+        - python                                 # [build_platform != target_platform]
+        - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+        - pybind11                               # [build_platform != target_platform]
+        - pyarrow 9.0.*                          # [build_platform != target_platform]
+        - numpy                                  # [build_platform != target_platform]
       host:
         - libtiledbsoma {{ version }}
         - pip
@@ -53,6 +65,8 @@ outputs:
         - scanpy
         - scipy
         - spdlog
+    imports:
+      - tiledbsoma
     test:
       commands:
         - python -c "import tiledbsoma; print(tiledbsoma.pytiledbsoma.version())"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,19 +1,17 @@
 {% set name = "tiledbsoma" %}
 {% set version = "1.1.1" %}
-{% set sha256 = "ac01357ed1f793fe96f728e4abb4965c7303918510cec38fda1ea5ef5ad47643" %}
+{% set sha256 = "a5ee104c326104d2a95777ca4b05f67d3422198879abc06f9269e6e5d101e037" %}
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 source:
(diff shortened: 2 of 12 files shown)


recipe/build-tiledbsoma-py.sh
```diff
@@ -4,4 +4,5 @@ set -ex
 
 cd apis/python
 
+echo "1.1.1" >> RELEASE-VERSION
```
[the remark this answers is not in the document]
nguyenv: https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/9/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR2

Yeah I agree that it would be better to not hardcode this. I wanted to utilize version from the `meta.yaml` file but couldn't figure out a way to do it.

recipe/build-tiledbsoma-py.sh
```diff
@@ -4,5 +4,5 @@ set -ex
 
 cd apis/python
 
-echo "1.1.1" >> RELEASE-VERSION
+echo "$PKG_VERSION" >> RELEASE-VERSION
```
johnkerl, line 7: first assert "$PKG_VERSION" is not empty -- ?

.ci_support/linux_64_.yaml
```diff
@@ -23,3 +29,6 @@ spdlog:
 - '1.11'
 target_platform:
 - linux-64
+zip_keys:
```
jdblischak, line 32: The numpy pins in `recipe/conda_build_config.yaml` aren't propagated because numpy is not in either the build or host requirements

conda-forge.yml
```diff
@@ -1 +1,14 @@
-{}
\ No newline at end of file
+# - only affects README installation instructions
+# - edit recipe/conda_build_config.yaml to affect build
+channels:
+  targets:
+    - ["tiledb", "main"]
+github:
+  user_or_org: TileDB-Inc
+  branch_name: master
```
jdblischak, line 8: The primary branch name for this feedstock repo is "main". It is only older feedstock repos like tiledb-vcf-feedstock that have a default branch of "master"

recipe/meta.yaml
```diff
@@ -1,45 +1,53 @@
...
     script: build-tiledbsoma-py.sh
     build:
-      binary_relocation: 
-        - lib/python*/site-packages/tiledbsoma/libtiledbsoma.*
-        - lib/python*/site-packages/tiledbsoma/pytiledbsoma.*.so
     requirements:
       build:
         - {{ compiler('cxx') }}
+        - python                                 # [build_platform != target_platform]
+        - cross-python_{{ target_platform }}     # [build_platform != target_platform]
```
jdblischak, line 47: What about pybind11? It is listed in `setup_requires` in `setup.py`, so I assume it should be required for cross-compilation

https://github.com/single-cell-data/TileDB-SOMA/blob/5eaa7e39d0ec116b6058ecc8bfecd207644d7b98/apis/python/setup.py#L248

And what about either numpy or pyarrow? Are these required during compilation? They are included in tiledb-vcf-feedstock:

https://github.com/TileDB-Inc/tiledb-vcf-feedstock/commit/aa0b77053441c78a8277adb23a43ed7480c2f265

recipe/meta.yaml
```diff
@@ -1,45 +1,53 @@
...
       host:
-        - libtiledbsoma {{ version }}
-        - pip
```
jdblischak, line 41: Is there a reason to remove `pip`? I think conda by default will pull in pip automatically with python unless a user disables it in their config file, but in generally `pip` is supposed to be listed in the host requirements

https://conda-forge.org/docs/maintainer/adding_pkgs.html#use-pip

recipe/meta.yaml
```diff
@@ -1,45 +1,53 @@
...
     build:
-      binary_relocation: 
-        - lib/python*/site-packages/tiledbsoma/libtiledbsoma.*
-        - lib/python*/site-packages/tiledbsoma/pytiledbsoma.*.so
     requirements:
       build:
         - {{ compiler('cxx') }}
+        - python                                 # [build_platform != target_platform]
+        - cross-python_{{ target_platform }}     # [build_platform != target_platform]
       host:
```
jdblischak, line 48: numpy needs to be added to the host requirements and then rerendered to use the numpy pins in `recipe/conda_build_config.yaml`. If pyarrow is also needed during compile time, it should also be added here

recipe/meta.yaml
```diff
@@ -49,10 +57,17 @@ outputs:
...
+        - pyarrow 9.0.*
         - scipy
-        - spdlog
+        - anndata
+        - tiledb-py 0.21.*
+        - typing-extensions
+        - numba
+        - attrs
+        - somacore 1.0.*
+        - scanpy 1.9.*
```
jdblischak, line 68: `python` needs to be explicitly added to the run requirements. That is why the current binaries of the Python client don't have build strings that start with `py`

recipe/meta.yaml
```diff
@@ -49,10 +57,17 @@ outputs:
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
         - anndata
         - pandas
-        - pyarrow
-        - scanpy
+        - pyarrow 9.0.*
         - scipy
-        - spdlog
+        - anndata
```
jdblischak, line 62: anndata is already listed above

jdblischak:
* The build strings now include `py` ([source](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=31195&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d&l=2620)):

```
  "tiledbsoma-py-1.1.2-py310h40e830b": {
    "recipe": {
      "channel_targets": "tiledb main",
      "cxx_compiler": "gxx",
      "cxx_compiler_version": "11",
      "numpy": "1.21",
      "python": "3.10.* *_cpython",
      "target_platform": "linux-64"
    }
  },
```

* The numpy pins are working. Python 3.7, 3.8, and 3.9 build with numpy 1.20, and Python 3.10 builds with numpy 1.21 (numpy 1.23 is installed in the test environments because of the `pin_compatible()` range in the run requirements)

jdblischak:
@Shelnutt2 Looks like there is a new dependency:

```
+ python -c 'import tiledbsoma; print(tiledbsoma.pytiledbsoma.version())'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/runner/miniforge3/conda-bld/tiledbsoma_1680000571072/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.8/site-packages/tiledbsoma/__init__.py", line 109, in <module>
    from ._general_utilities import (
  File "/Users/runner/miniforge3/conda-bld/tiledbsoma_1680000571072/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.8/site-packages/tiledbsoma/_general_utilities.py", line 11, in <module>
    import distro
ModuleNotFoundError: No module named 'distro'
```

The only reason the osx-arm64 build passed was because it skips the tests

Turns out `distro` was added yesterday: https://github.com/single-cell-data/TileDB-SOMA/commit/5eaa7e39d0ec116b6058ecc8bfecd207644d7b98

PR #13, opened 2023-04-13
Update recipe for new `TILEDBSOMA_PATH` handling

johnkerl:
As pointed out by @jdblischak on #12 with regard to changes made on https://github.com/single-cell-data/TileDB-SOMA/pull/1189.

### DIFF
--- a/recipe/build-tiledbsoma-py.sh
+++ b/recipe/build-tiledbsoma-py.sh
@@ -5,4 +5,5 @@ set -ex
 cd apis/python
 
 echo "$PKG_VERSION" >> RELEASE-VERSION
-$PYTHON setup.py install --single-version-externally-managed --record record.txt --libtiledbsoma="${PREFIX}"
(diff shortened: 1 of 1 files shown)


jdblischak:
I'm pretty sure this change needs to be made after the next release of TileDB-SOMA, right? IIUC the current release of the Python client doesn't support `TILEDBSOMA_PATH`.

Also, I thought the plan under consideration was to allow the Python client to support both strategies, ie the user could set `TILEDBSOMA_PATH` or use `--libtiledbsoma`, in which case the conda recipe wouldn't need to be udpated

johnkerl:
Thanks for the feedback @jdblischak ! :)

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/pull/1253 FYI

johnkerl:
Superseded by https://github.com/single-cell-data/TileDB-SOMA/pull/1253

PR #17, opened 2023-04-26
Depend on core >= 2.15.2, < 2.16

johnkerl:
As in https://github.com/single-cell-data/TileDB-SOMA/pull/1296 and https://github.com/single-cell-data/TileDB-SOMA/pull/1303, for the same reasons already narrated there.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,9 +23,10 @@ source:
 
 
 build:
-  number: 0
+  number: 1
   skip: true  # [win or linux32 or py2k]
 # Important: set this back to 0 on a new release
+# Important: increment if there is not a new release
 
 outputs:
   - name: libtiledbsoma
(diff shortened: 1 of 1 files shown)


Shelnutt2 (changes requested):
Pinning needs to have an upperbound to prevent bringing in 2.16 or newer. Something like `tiledb >=2.15.1,<2.16` (taken from https://github.com/conda-forge/tiledb-py-feedstock/blob/f153315c81da2ed5f2d89a968c94892aca4b0bde/recipe/meta.yaml#L35)

jdblischak:
Since you are changing the recipe but not the package version, now is when you bump the build number from `0` to `1`

johnkerl:
@jdblischak @Shelnutt2 thank you both for teaching me things!

That said, I'm going to hold off b/c TileDB-Py 0.21.3 with core 2.15.2 is just around the corner:
https://github.com/TileDB-Inc/TileDB-Py/pull/1754

johnkerl:
Folding this into #18

PR #22, opened 2023-05-05
Add recipe output for r-tiledbsoma

jdblischak:
This PR adds an output section to build the conda package r-tiledbsoma

* Good news: I was able to build the recipe locally on Ubuntu for R 4.1 and 4.2. I've copied some relevant output from the successful build below

* Bad news: The Azure builds fail. For some strange reason, the conda solver tries to install R 3.5. Still have to investigate

* Temporary changes: note that I made a few temporary changes to facilitate testing. First I removed all but one of the Python variants to reduce the build time. Second I switched to use the Git repo as the source since enabling the conda builds is an ongoing process

* Immediate question: The Python client version exactly matches the libtiledbsoma version. Is the plan to keep a separate version string for the R client?

cc: @eddelbuettel, @nguyenv, @aaronwolen, @ihnorton 
xref: https://github.com/single-cell-data/TileDB-SOMA/issues/1277#issuecomment-1520930305, https://github.com/single-cell-data/TileDB-SOMA/pull/1302

Excerpts from my successful local build:

```sh
> tiledbsoma::show_package_versions()
tiledbsoma:    0.0.0.9022
tiledb-r:      0.19.1
tiledb core:   2.15.2
libtiledbsoma: libtiledb=2.15.2
R:             R version 4.1.3 (2022-03-10)
OS:            Ubuntu 22.04.2 LTS

> tiledbsoma::show_package_versions()
tiledbsoma:    0.0.0.9022
tiledb-r:      0.19.1
tiledb core:   2.15.2
libtiledbsoma: libtiledb=2.15.2
R:             R version 4.2.3 (2023-03-15)
OS:            Ubuntu 22.04.2 LTS
```

```json
{
  "libtiledbsoma-1.2.2-h238ffd6_0": {
    "recipe": {
      "c_compiler": "gcc",
      "c_compiler_version": "12",
      "channel_targets": "conda-forge main",
      "cxx_compiler": "gxx",
      "cxx_compiler_version": "12",
      "fmt": "9",
      "spdlog": "1.11",
      "target_platform": "linux-64"
    }
  },
  "r-tiledbsoma-0.0.0.9022-r41h00ab1b0_0": {
    "recipe": {
      "channel_targets": "conda-forge main",
      "cxx_compiler": "gxx",
      "cxx_compiler_version": "12",
      "target_platform": "linux-64"
    }
  },
  "r-tiledbsoma-0.0.0.9022-r42h00ab1b0_0": {
    "recipe": {
[...]

PR #21, opened 2023-05-03
TileDB-SOMA 1.2.3

johnkerl:

PR #20, opened 2023-05-03
Troubleshoot solver error

jdblischak:
The recipe that successfully built last week (bb3fbeb622d88809cf20bc8b58c3a480803334c7) is now failing. Specifically, libtiledbsoma builds fine, but then the Python client build fails because conda isn't able to solve the environment:

```sh
- package libtiledbsoma-1.2.2-h25619bb_0 requires tiledb >=2.15.2,<2.16.0a0, but none of the providers can be installed
```

Some incompatibility has arisen somewhere in the dependency chain. I'm using this PR to troubleshoot

xref: #18

PR #23, opened 2023-05-22
Pre-tag check for TileDB-SOMA 1.2.4

johnkerl:
This is _not_ the release PR -- this is a pre-release check PR following our standard procedure at

### DIFF
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -1,7 +1,7 @@
 c_compiler:
 - gcc
 c_compiler_version:
-- '11'
+- '12'
 cdt_name:
 - cos6
 channel_sources:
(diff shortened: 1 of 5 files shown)


jdblischak:
Do you still need to update the pins for `tiledb` and `tiledb-py` like you had done in PR #21?

johnkerl:
Good call -- thanks @jdblischak !

jdblischak:
The linux build failed with a conda solver error when creating the environment to build the Python env. It's the same solver error we battled in #20 that only affects Python 3.10. I was hoping it was only transient. I had to switch from testing Python 3.10 to Python 3.9 last week when testing PR #22 (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/22/commits/75120f788df8fb9d1477a96672a30364ee37146d)

jdblischak:
Also, I recommend rerendering to see if that helps `conda smithy rerender --commit auto`

PR #26, opened 2023-06-06
TileDB-SOMA 1.2.5

johnkerl:
#24 gets us TileDB-SOMA 1.2.4: https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.2.4

Latest tagged in GitHub, and in [PyPI](https://pypi.org/project/tiledbsoma/#history), is 1.2.5: https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.2.5

So this PR will get PyPI and Conda deliverables both on latest-released

The only dependency update here is `somacore` 1.0.2 -> 1.0.3 but that's handled by
https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/main/recipe/meta.yaml#L103

Tracking: https://github.com/single-cell-data/TileDB-SOMA/issues/1455

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -100,7 +100,7 @@ outputs:
         - typing-extensions >=4.1
         - numba
         - attrs
-        - somacore 1.0.*
+        - somacore >=1.0.3
         - scanpy 1.9.*
     test:
       imports:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.2.4" %}
+{% set version = "1.2.5" %}
 {% set version_r = "0.0.0.9024" %}
-{% set sha256 = "3b1cdfb5b1a9572c191c30cbab8c11d2344d3d6015dbc45cb4e66eca8ab2eb54" %}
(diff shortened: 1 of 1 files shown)


jdblischak:
> The only dependency update here is somacore 1.0.2 -> 1.0.3 but that's handled by

How strong is that dependency? Could tiledbsoma-py run with somacore 1.0.2? That would be an allowable configuration according to the conda recipe because it only pins to major.minor (`somacore 1.0.*`), whereas `setup.py` pins to `major.minor.patch` (`"somacore==1.0.3"`)

https://github.com/single-cell-data/TileDB-SOMA/blob/7ade0b7cfe59c75c35f4d8e8bd8586404cae1abd/apis/python/setup.py#L291

johnkerl:
@jdblischak it's important -- `somacore` 1.0.3 is part of the deliverables on `tiledbsoma-py` 1.2.5

I suppose we need a minimum-pin on `somacore` ...

jdblischak:
somacore 1.0.3 isn't available as a conda binary:

> nothing provides somacore >=1.0.3 needed by tiledbsoma-py-1.2.5-py37h0f1fff0_0

Looks like the only somacore release that was built for conda was 1.0.0

https://github.com/TileDB-Inc/somacore-feedstock
https://anaconda.org/tiledb/somacore/files

jdblischak (approved):
I confirmed locally that rerendering isn't required

PR #29, opened 2023-06-15
Fix Azure links in README

jdblischak:
Just like in https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/86, this is a documentation-only update to fix the links to the latest Azure build in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,8 @@ Current build status
     <td>
       <details>
         <summary>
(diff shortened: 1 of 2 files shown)


jdblischak:
Unfortunately there is no way to prevent the needless CI jobs for a PR. However, with `[ci skip]` in the commit message title, the jobs won't run once again after merging (I won't create a merge commit)

johnkerl (approved):

PR #30, opened 2023-06-20
TileDB-SOMA 1.2.6 pre-release check

johnkerl:
Following our currently established procedure at

As always (please see the above wiki) this isn't to _be_ released but rather to check the _feasibility_ of tagging a release.

Food for thought: The four PRs at the tip of https://github.com/single-cell-data/TileDB-SOMA/commits/release-1.2 as of today are all app-level -- and as such shouldn't affect package pins -- _except_ for https://github.com/single-cell-data/TileDB-SOMA/pull/1484/files. That PR affects the `numba` pin, only for Python 3.11: from 0.57.0rc1 to 0.57.0. However, in this repo's `recipe/meta.yaml` we have no `numba` constraints so I don't understand what we do or do not need to do new here.

Another note: I haven't re-renedered yet but I will attempt to check to see if we need to re-render.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.2.5" %}
+{% set version = "1.2.6" %}
 {% set version_r = "0.0.0.9024" %}
-{% set sha256 = "0eb5b44f0a786d3089fb4ce7ee5d5c4ee6df1c7cacaa8a77f11bde95c1b244ce" %}
+{% set sha256 = "this-is-tbd-see-below" %}
 # This is the SHA256 of
 #   TileDB-SOMA-i.j.k.tar.gz
 # from
@@ -13,19 +13,19 @@ package:
   name: {{ name }}
   version: {{ version }}
 
-source:
-  url: https://github.com/single-cell-data/TileDB-SOMA/archive/{{ version }}.tar.gz
-  sha256: {{ sha256 }}
-# Pre-release canary "will Conda be green if we release":
 #source:
-#  git_url: https://github.com/single-cell-data/TileDB-SOMA.git
-#  git_rev: db2782bad8b11f7eb0899d0a8b78118c6b5a49a2
-#  git_depth: 1
-#  # hoping to be 1.2.4
+#  url: https://github.com/single-cell-data/TileDB-SOMA/archive/{{ version }}.tar.gz
+#  sha256: {{ sha256 }}
+# Pre-release canary "will Conda be green if we release":
+source:
+  git_url: https://github.com/single-cell-data/TileDB-SOMA.git
+  git_rev: 1c085c51eef6f7b849e6cd3def151afbc5a5ca47 # release-1.2 tip as of 2023-06-20
+  git_depth: 1
+  # hoping to be 1.2.6
 
 
 build:
-  number: 1
+  number: 0
   skip: true  # [win or linux32 or py2k]
 # Important: set this back to 0 on a new release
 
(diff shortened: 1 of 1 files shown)


johnkerl:
TIL @jdblischak -- thanks -- always one more thing to put into my runbook!

jdblischak:
> That PR affects the numba pin, only for Python 3.11: from 0.57.0rc1 to 0.57.0. However, in this repo's recipe/meta.yaml we have no numba constraints so I don't understand what we do or do not need to do new here.

I had previously raised concern about the the lack of a numba pin. However, I remembered that currently tiledbsoma-py is only built for Python 3.7, 3.8, 3.9, and 3.10

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/7610f021f4b7c7d09b5934f84a89bdbaf04404c6/recipe/conda_build_config.yaml#L16-L21

And even conda-forge doesn't yet build for Python 3.11

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/28852e81a79eeb6f9ab63370647e582f8769021b/recipe/conda_build_config.yaml#L669-L674

I skimmed  https://github.com/single-cell-data/TileDB-SOMA/issues/1463 and https://github.com/single-cell-data/TileDB-SOMA/pull/1472, but the constraints are not entirely clear to me. Python 3.11 appears to require numba 0.57, which we don't need to worry about in this recipe until we start building binaries for Python 3.11. But what about Python 3.7, 3.8, 3.9, and 3.10? Can they really not work with numba 0.57? If that's the case, then yes, we should add a pin such as `numba <0.57`

johnkerl:
> Can they really not work with numba 0.57?

@jdblischak that's not the reason for the 0.57 pin for Python 3.11 only

It was the other way around, in fact -- older numba wouldn't work with Python 3.11, and in fact, we couldn't support `tiledbsoma-py` on Python 3.11 at all until numba 0.57.0rc1 came out

Links:

* https://github.com/single-cell-data/TileDB-SOMA/issues/544
* https://github.com/single-cell-data/TileDB-SOMA/issues/1463

If we aren't doing Python 3.11 in this feedstock repo, and if only Python 3.11 leads to the `numba==0.57` constraint, then I'm in favor of _not_ adding any numba constraints right now. (Solving is hard, and the more pins we add, the harder it gets.)

jdblischak:
Looking at the [release notes for numba 0.57](https://numba.readthedocs.io/en/stable/release-notes.html#version-0-57-0-1-may-2023), my understanding is that numba 0.57 is the first to add support for Python 3.11, and that the minimum required version was bumped to Python 3.8. Thus for the purposes of this recipe, we could update the `numba` pin to enforce that Python 3.7 requires numba < 0.57. However, one could reasonably argue that this is overkill. Since numba 0.57 dropped support for Python 3.7, conda-forge obviously isn't building a binary for this combination (in fact it has dropped automatic support for Python 3.7, and feedstocks have to opt-in to continue building for Python 3.7 like we have for tiledbsoma-py).

jdblischak:
> If we aren't doing Python 3.11 in this feedstock repo, and if only Python 3.11 leads to the numba==0.57 constraint, then I'm in favor of not adding any numba constraints right now. (Solving is hard, and the more pins we add, the harder it gets.)

@johnkerl agreed

jdblischak:
> Another note: I haven't re-renedered yet but I will attempt to check to see if we need to re-render.

I rerendered locally to test. The only update was dropping the R 4.1 build. As long at the R 4.1 build continues to succeed, I see no rush to remove it (and thus recommend skipping rerendering for this release)

jdblischak:
Also, I confirmed in the linux-64 build that the locally built libtiledbsoma was always installed when solving the environments to build the API clients (ie the recent addition of `{{ pin_subpackage('libtiledbsoma', exact=True) }}` to the host requirements is continuing to work as expected)

jdblischak (approved):

PR #31, opened 2023-06-21
Revert #30

johnkerl:
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/30 was a trial balloon only and yet I merged it mistakenly

Please see

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,7 +1,7 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.2.6" %}
+{% set version = "1.2.5" %}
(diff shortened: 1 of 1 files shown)


jdblischak (approved):

jdblischak:
Also, for documentation purposes: @ihnorton canceled the [Azure builds](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=32765&view=results) and then manually deleted the linux-64 binaries that had already been uploaded: 1 libtiledbsoma; 4 tiledbsoma-py (1 for each Python version); 0 r-tiledbsoma (1.2.6 is a Python-only fix, so the R version was unchanged)

PR #32, opened 2023-06-21
TileDB-SOMA 1.2.7

johnkerl:
Following
and with pre-release green-light check from #30

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -25,7 +25,7 @@ source:
 
 
 build:
-  number: 1
+  number: 0
   skip: true  # [win or linux32 or py2k]
 # Important: set this back to 0 on a new release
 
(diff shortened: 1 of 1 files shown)


jdblischak:
Documenting our scattered discussions here:

* No need to bump `version_r` because no change has been made to the R client for this patch release
* No need to rerender. We already tried it locally, and the only change is dropping the build for R 4.1 (conda-forge is migrating from 4.1 to 4.3). This release doesn't even generate new R binaries (because of the previous point), so it wouldn't have an immediate effect. And as long as the R 4.1 build is passing, there is no rush to remove support
* The patch version is 7 instead of 6 (as it was in #30) to avoid any potential confusion (see #31)

jdblischak (changes requested):
Bump the build number back to `0` like you had done in #30. This will be the first build of version 1.2.7

jdblischak (approved):

PR #38, opened 2023-08-04
Always clean up Python and C++ builds

Shelnutt2:
This ensures in a multi-package build scenario each source build starts with a clean directory.

This is a further of the initial work that cleaned up the R package after each build.

Note: I didn't bump the build number because this won't cause any artifact changes and we don't need to re-upload.

### DIFF
--- a/recipe/build-libtiledbsoma.sh
+++ b/recipe/build-libtiledbsoma.sh
@@ -15,3 +15,5 @@ cmake \
 make -j ${CPU_COUNT}
 
 make install-libtiledbsoma
+
+cd .. && rm -rf libtiledbsoma-build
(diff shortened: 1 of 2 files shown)


johnkerl (approved):

jdblischak (approved):

jdblischak:
My only question is: why didn't this bite us before? It's my understanding that the Python client has compiled code just like the R client. Doesn't this mean that the compiled code included in all the Python binaries is from whatever Python variant happened to be built first?

jdblischak:
> My only question is: why didn't this bite us before? It's my understanding that the Python client has compiled code just like the R client. Doesn't this mean that the compiled code included in all the Python binaries is from whatever Python variant happened to be built first?

I figured out the answer to my question. The build directory for each Python variant includes the Python version, eg:

```sh
creating build/lib.linux-x86_64-cpython-39
creating build/lib.linux-x86_64-cpython-39/tiledbsoma
copying src/tiledbsoma/_tdb_handles.py -> build/lib.linux-x86_64-cpython-39/tiledbsoma
```

And currently there is only a single variant for the C++ build.

> Note: I didn't bump the build number because this won't cause any artifact changes and we don't need to re-upload.

Thus I agree that the artifacts won't change. And I even downloaded a few binaries to spot check, and I confirmed that only the files for the given version of Python are packaged.

However, to be on the safe side, especially if we start building multiple C++ variants, I think we should merge this PR as is

PR #36, opened 2023-08-04
tiledbsoma 1.4.0 test run

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SOMA/issues/1567

See also https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

PR #39, opened 2023-08-05
tiledbsoma 1.4.0

johnkerl:
Following

* https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/36
* https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -42,11 +42,11 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        - tiledb >=2.16.1,<2.17
+        - tiledb >=2.16.2,<2.17
         - spdlog
         - fmt
       run:
-        - tiledb >=2.16.1,<2.17
+        - tiledb >=2.16.2,<2.17
         - spdlog
         - fmt
     about:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,5 @@
 {% set name = "tiledbsoma" %}
 {% set version = "1.4.0" %}
-{% set version_r = "1.4.0" %}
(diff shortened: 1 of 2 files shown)


johnkerl:
"Queued" for 3+ hours; not sure what's going on ...

recipe/meta.yaml
```diff
@@ -13,20 +13,20 @@ package:
   name: {{ name }}
   version: {{ version }}
 
-source:
-  url: https://github.com/single-cell-data/TileDB-SOMA/archive/{{ version }}.tar.gz
-  sha256: {{ sha256 }}
+#source:
```
Shelnutt2, line 16: This section for source + url + sha256 needs to be uncommented.

recipe/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.3.0" %}
-{% set version_r = "0.99.1" %}
-{% set sha256 = "1fb5bf86440a47b724ffb083143bda8a26d79e036336bac57ac7c1feaa5d4b9d" %}
+{% set version = "1.4.0" %}
+{% set version_r = "1.4.0" %}
```
Shelnutt2, line 3: Shall we replace the `version_r` and just use `version` now?

Shelnutt2 (changes requested):

recipe/meta.yaml
johnkerl: yes; good call definitely

recipe/meta.yaml
johnkerl: that was an embarrassingly dumb mistake on my part 😐

johnkerl:
@Shelnutt2 all checks are passing

Shelnutt2 (approved):

PR #47, opened 2023-09-20
Always update conda-smithy for nightly job

jdblischak:
conda-smithy checks if it is the latest version available, and since the job caches the conda env, it fails to rerender the feedstock every time there is a new conda-smithy release


johnkerl (approved):
🚢

PR #54, opened 2023-11-02
tiledbsoma 1.5.0rc3

johnkerl:
I am starting this rolling in case of any delay with 1.5.0 -- if 1.5.0 comes out soon I'll cancel this

Established procedure:


jdblischak:
Does tiledbsoma 1.5.0 rely on the new API introduced in [libtiledb 2.17.3](https://github.com/TileDB-Inc/TileDB/releases/tag/2.17.3)? If yes, the tiledb pins need to be updated to `tiledb >=2.17.3,<2.18`

jdblischak (approved):

PR #63, opened 2023-11-17
Update pyarrow dependency

johnkerl:
Only a WIP canary to check satisfiability

https://github.com/single-cell-data/TileDB-SOMA/issues/1926

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -72,7 +72,7 @@ outputs:
         - python                                 # [build_platform != target_platform]
         - cross-python_{{ target_platform }}     # [build_platform != target_platform]
         - pybind11                               # [build_platform != target_platform]
-        - pyarrow                                # [build_platform != target_platform]
+        - pyarrow!=14.0.0                        # [build_platform != target_platform]
         - numpy                                  # [build_platform != target_platform]
       host:
         - python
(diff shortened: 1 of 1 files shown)


jdblischak (approved):
Now that pyarrow 14.0.1 has been released, it is unlikely that 14.0.0 would ever be installed. But I am fine with be cautious (ie better safe than sorry)

caveat: I am only approving the strategy. We would need to bump the build number before merging

johnkerl:
@jdblischak this is in draft and incomplete. I misread the notification. Affected ranges are 0.14.0 to 14.0.0. So I need to modify the pin. Also, pyarrow 14 doesn't work with Python 3.7. @thetorpedodog pointed out a hotfix option elsewhere. I need to circle back to this today after meetings.

Also: @bkmartinjr has already verified that `tiledbsoma` does _not_ use the codepath within `pyarrow` that's vulnerable. So, our bumping our dependency is a good idea, and good optics, but not drop-everything-do-it-now.

johnkerl:
> caveat: I am only approving the strategy. We would need to bump the build number before merging

@jdblischak I think I want this as a trial balloon only, just to check for buildability, and not try to merge until tiledbsoma 1.6.0 which will be happening very soon, as core 2.18.0 has already been tagged.

johnkerl:
Commit https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/63/commits/21fe23d54f9a7c602f2cf77fafd80c72892ba89a implements the plan at https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1821208184

johnkerl:
```
ubuntuip-172-31-85-136[tiledbsoma-feedstock] conda smithy rerender --commit auto
...
INFO:conda_smithy.configure_feedstock:No changes made. This feedstock is up-to-date.
```

jdblischak (approved):

johnkerl:
Closing as this was just a pre-tag canary -- thank you @jdblischak !

PR #64, opened 2023-12-01
tiledbsoma 1.6.0 pre-release check 1

johnkerl:
As usual: https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases

except: https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1834695149

-- here we need to validate:

* core 2.18
* `tiledb-py` 0.24
* `tiledb-r` 0.22
* `pyarrow` as noted in the linked issue above

There will be a `somacore` bump later, and another `tiledbsoma` PR, before 1.6.0 -- so there will be a second Conda pre-release-check PR, following on this one

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -73,6 +73,7 @@ outputs:
         - cross-python_{{ target_platform }}     # [build_platform != target_platform]
         - pybind11                               # [build_platform != target_platform]
         - pyarrow                                # [build_platform != target_platform]
+        - pyarrow_hotfix                         # [build_platform != target_platform]
         - numpy                                  # [build_platform != target_platform]
       host:
         - python
@@ -89,6 +90,7 @@ outputs:
         # https://github.com/single-cell-data/TileDB-SOMA/issues/1926
         - pyarrow <13.0 # [osx]
         - pyarrow       # [not osx]
+        - pyarrow_hotfix
       run:
         - {{ pin_compatible('numpy', lower_bound='1.16', upper_bound='1.24') }}
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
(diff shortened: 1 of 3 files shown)


recipe/meta.yaml
```diff
@@ -86,17 +89,23 @@ outputs:
         - setuptools_scm
         - wheel
         - numpy
-        - pyarrow
+        # https://github.com/single-cell-data/TileDB-SOMA/issues/1926
+        - pyarrow <13.0 # [osx]
```
jdblischak, line 93: I suspect that the error loading S3 objects described in https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1834695149 is a runtime-only error. However, given the complexity of the [libarrow run_exports](https://github.com/conda-forge/arrow-cpp-feedstock/blob/62c529770482209a33e3efd93c16a8779eda181f/recipe/meta.yaml#L162), I agree it is safest to also enforce these pins at build time

jdblischak (approved):

recipe/meta.yaml
johnkerl: @jdblischak my thoughts exactly

jdblischak:
I checked the build logs to confirm that:

1. pyarrow < 13 was installed for the osx and osx-arm64 builds (typically pyarrow 11 or 12 was installed)
2. pyarrow-hotfix was installed for all the builds

Interestingly, for the linux builds (except py37), tiledbsoma-py was built against pyarrow <13, but then pyarrow 14.0.1 was installed at runtime. Thus the libarrow run_exports constraints might not be as strict as we had thought. My guess is that since we depend on pyarrow (and not directly on libarrow), that its `max_pin = "x"` doesn't apply to tiledbsoma-py (to be clear, I don't think we need to change anything, just documenting the builds)

PR #69, opened 2023-12-13
tiledbsoma 1.5.2 pre-check

johnkerl:
Following our established procedure at

with the exception that TileDB-SOMA latest is 1.6.1 (see also #68) so here we are going to merge to the `release-1.5` branch of this repo, on the advice of @ihnorton

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -7,7 +7,7 @@ build_platform:
 conda_forge_output_validation: true
 test_on_native_only: true
 # build_with_mambabuild: false
-upload_on_branch: main
+upload_on_branch: release-1.5
 azure:
   build_id: 43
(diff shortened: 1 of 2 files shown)


johnkerl:
This pre-check PR has green CI, fulfilling its purpose

conda-forge.yml
```diff
@@ -7,7 +7,7 @@ build_platform:
 conda_forge_output_validation: true
 test_on_native_only: true
 # build_with_mambabuild: false
-upload_on_branch: main
+upload_on_branch: release-1.5
```
jdblischak, line 10: It doesn't matter for this pre-check, but a reminder that you will need to rerender for this edit in `conda-forge.yml` to take effect in the follow-up PR that will be merged

PR #74, opened 2024-01-09
tiledbsoma 1.6.2 pre-check

johnkerl:
Following our established procedure:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,13 @@
-*.pyc
+# User content belongs under recipe/.
+# Feedstock configuration goes in `conda-forge.yml`
+# Everything else is managed by the conda-smithy rerender process.
+# Please do not modify
 
-build_artifacts
+*
+!/conda-forge.yml
+
+!/*/
+!/recipe/**
+!/.ci_support/**
+
+*.pyc
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -42,11 +42,11 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        - tiledb >=2.18.2,<2.19
+        - tiledb =2.18.2
         - spdlog
         - fmt
       run:
-        - tiledb >=2.18.2,<2.19
+        - tiledb =2.18.2
         - spdlog
         - fmt
     about:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -45,10 +45,6 @@ outputs:
         - tiledb =2.18.2
         - spdlog
         - fmt
-      run:
-        - tiledb =2.18.2
-        - spdlog
-        - fmt
     about:
       home: http://tiledb.com
       license: MIT
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.6.1" %}
-{% set sha256 = "0ef61bc6e5922cb68e62f443c353181f29f7ff7def79dbbf8a030897e2de8a38" %}
+{% set version = "1.6.2" %}
+{% set sha256 = "tee-bee-dee" %}
 # This is the SHA256 of
 #   TileDB-SOMA-i.j.k.tar.gz
 # from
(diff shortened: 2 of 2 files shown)


johnkerl:
I do not understand
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/74/checks?check_run_id=20322245882
-- this attempted at a 1.6.2 is a bugfix-PR delta from [1.6.1](https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.6.1) and I don't see why it won't solve

Specifically, the Azure error message is
-   - package libtiledbsoma-1.6.2-hd763302_0 requires tiledb >=2.18.3,<2.19.0a0, but none of the providers can be installed
but I don't see why this does not solve given that 1.6.1 did solve (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/68) ...

johnkerl:
Additionally confusing: the bit `requires tiledb >=2.18.3` given that this PR only has `tiledb >=2.18.2,<2.19` ... 🤔

jdblischak:
> Additionally confusing: the bit requires tiledb >=2.18.3 given that this PR only has tiledb >=2.18.2,<2.19

This is because of the ["run exports" of the tiledb recipe](https://github.com/conda-forge/tiledb-feedstock/blob/5626a0dad734250ef17a6ca9bed0cb9d1fd2feee/recipe/meta.yaml#L17C52-L17C52). libtiledbsoma was built against libtiledb 2.18.3, so all the downstream clients have to install libtiledb 2.18.3 or greater as well

> but I don't see why this does not solve given that 1.6.1 did solve (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/68) ...

libtiledbsoma 1.6.1 was built against libtiledb 2.18.2

The conda solver issues have been brutal lately. These solver errors are the same problem that plagued the recent release of TileDB-VCF (https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/105). There is some conflict between the libraries linked to libtiledb 2.18.3 and the libraries linked to something needed for the Python builds. I suspect libarrow/pyarrow, but I haven't been able to diagnose nor fix the solver issue despite trying multiple ideas last month

A short-term solution would be to pin to exactly `tiledb 2.18.2`, which should solve just like the 1.6.1 release did

johnkerl:
@jdblischak thank you! https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/74/commits/ae9085e3dce7575384dafcf2108c94b045887b8a 🤞

jdblischak:
The solver was failing because the tiledbsoma-py builds (except py37) and r-tiledbsoma builds require different versions of libtiledb:

* libtiledb 2.18.2: tiledbsoma-py (py38, py39, py310)
* libtiledb 2.18.3: tiledbsoma-py (py37 - linux-only), r-tiledbsoma (R 4.2 & 4.3)

Thus to enable flexibility in the patch version of libtiledb that is installed in the downstream clients, we had to do the following:

1. Build libtiledbsoma against libtiledb 2.18.2. This results in the runtime requirement of `tiledb >=2.18.2,<2.19.0a0`, which is the behavior of `max_pin="x.x"` in the "run exports" of the libtiledb recipe
2. Remove the strict run requirement of libtiledbsoma. This allows it to be co-installed with either libtiledb 2.18.2 or 2.18.3 as required by the downstream client

As I documented in https://github.com/conda-forge/tiledb-feedstock/pull/215, software built against a previous minor version of libtiledb (eg 2.18.2) is compatible at runtime with later patch releases (eg 2.18.3). The converse is not true. Software built against a later version (eg 2.18.3) is not guaranteed to be compatible at runtime with previous patch releases (eg 2.18.2). I learned this the hard way with the new API introduced in libtiledb [2.17.3](https://github.com/TileDB-Inc/TileDB/releases/tag/2.17.3)

jdblischak:
And a note to my future self or anyone else trying to fix conda solver issues in the future:

When I render the recipe locally, the `tiledb 2.18.3` pin is inserted into the host requirement section of r-tiledbsoma. This is strange since `tiledb` is not a top-level requirement of r-tiledbsoma. The tiledbsoma-py recipe behaves as expected, ie only libtiledbsoma and tiledb-py are in the host requirements. If we can fix this behavior, this should make future tiledbsoma-feedstock updates easier

johnkerl:
This pre-check PR has done its job following our established procedure:

PR #76, opened 2024-01-18
tiledbsoma 1.6.2 with core 2.18.4

johnkerl:


jdblischak:
I don't think this PR is necessary. The tiledb run exports are `"x.x"`, which means a tiledbsoma binary built against tiledb 2.18.x can't be installed with any of tiledb version >=2.18.x,<2.19

```sh
mamba create --dry-run -n test \
  -c conda-forge -c tiledb --override-channels \
  tiledb=2.18.4 tiledbsoma-py=1.6.2
## + azure-core-cpp                  1.10.3  h91d86a7_0               conda-forge     Cached
## + tiledb                          2.18.4  h4386cac_0               conda-forge        6MB
## + libtiledbsoma                    1.6.2  hd763302_0               tiledb           343kB
## + python                         3.10.13  hd12c33a_1_cpython       conda-forge     Cached
## + tiledbsoma-py                    1.6.2  py310h0107f82_0          tiledb           370kB
```

recipe/meta.yaml
```diff
@@ -42,7 +42,7 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        - tiledb =2.18.2
+        - tiledb 2.18.*
```
jdblischak, line 45: Recall that we purposefully built libtiledbsoma against libtiledb 2.18.2 so that the conda solver could be flexible about which libtiledb it installed when building the R and Python variants

https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/74#issuecomment-1885371268

If you switch back to `2.18.*`, this will build libtiledbsoma against 2.18.4, and force 2.18.4 to be installed for all the downstream builds (which appear to not solve)

johnkerl: @jdblischak and cc @ihnorton we definitely need to use 2.18.4 either via `2.18.*` or via `2.18.4` ...

johnkerl:
> I don't think this PR is necessary. The tiledb run exports are "x.x", which means a tiledbsoma binary built against tiledb 2.18.x can't be installed with any of tiledb version >=2.18.x,<2.19

@jdblischak I'm not certain either. I had thought notebook-images and UDF-image repos would fail to solve with this 2.18.2 pin here, when we need to be on 2.18.4 everywhere.

There's a parallel conversation in Slack right now; I thought I'd kick this off. But meanwhile we have attempted solves in flight for notebook and UDF images ... 🤞

recipe/meta.yaml
johnkerl: OK thanks @jdblischak !! I hadn't grasped correctly; thanks for straightening me out :)

johnkerl:
Per @jdblischak I am closing this PR

PR #48, opened 2023-09-28
Canary to check for `numba>=0.58`

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/pull/1723 made me nervous so I want to run a condability check

This is neither a release tag nor immediate prep for a release tag. But let's make this change before TileDB-SOMA 1.7.0.

Note numba 0.58.* isn't available for Python 3.7 and never will be as Anaconda has dropped support for Python 3.7.

PR #77, opened 2024-01-22
tiledbsoma 1.7.0 pre-check

johnkerl:
This follows our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases).

As well it includes (as it must) https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/48

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.6.2" %}
-{% set sha256 = "d9033f801c2d7f1cc9cf32e9d02ee8eaca88fe06b64b8073a3175f1f8d8f7c8d" %}
+{% set version = "1.7.0" %}
+{% set sha256 = "tee-bee-dee" %}
 # This is the SHA256 of
 #   TileDB-SOMA-i.j.k.tar.gz
 # from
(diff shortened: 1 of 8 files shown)


jdblischak (approved):

jdblischak:
With the update to tiledb 2.19, we can no longer hide from the solver problems that started with 2.18.3 and 2.18.4. I'll try to find time to investigate further later today

recipe/meta.yaml
```diff
@@ -102,9 +105,10 @@ outputs:
         - scipy
         - anndata <0.10 # [py>37]
         - anndata <0.9  # [py<=37]
-        - tiledb-py >=0.24.0,<0.25.0
+        - tiledb-py >=0.25.0,<0.26.0
         - typing-extensions >=4.1
-        - numba
+        - numba >=0.58.1 # [py>37]
+        - numba          # [py<=37]
```
jdblischak, line 111: I confirmed that this matches the upstream requirements in `setup.py`

https://github.com/single-cell-data/TileDB-SOMA/blob/5dac84866c5dfc74f0ff4f9b891c9b26902846c7/apis/python/setup.py#L283

Note that `~=0.58.0` is actually stricter than `>=`. The `~=` translates to a pin of `0.58.*`. However, I agree with the choice of `>=` for the conda recipe. It's better to allow more flexibility for the conda solver until we know for certain there is a limitation in the version range. We can always submit a repodata patch in the future to fix the metadata if a hypothetical numba 0.59 were to cause problems

jdblischak:
I haven't been able to figure out why, but for some reason, starting with tiledb 2.18.3, there is some conflict in the dependency solver for the tiledbsoma-py conda envs. The R build envs continue to be fine (they previously installed with 2.18.3, and are now happy to install with 2.19.0).

The build order this time started with the R builds, which both passed. The failure comes from the first Python build, 3.8:

```
BUILD START: ['libtiledbsoma-1.7.0-hd763302_0.conda', 'r-tiledbsoma-1.7.0-r43h415b3ff_0.conda', 'r-tiledbsoma-1.7.0-r42h415b3ff_0.conda', 'tiledbsoma-py-1.7.0-py38h362c0dd_0.conda', 'tiledbsoma-py-1.7.0-py310h0107f82_0.conda', 'tiledbsoma-py-1.7.0-py37h0f1fff0_0.conda', 'tiledbsoma-py-1.7.0-py39h580bbbf_0.conda']
```

Based on previous experience (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/74#issuecomment-1885371268), I assume this conda solver problem will affect the Python 3.8, 3.9, and 3.10 builds

johnkerl:
Following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) this PR has done its job.

jdblischak:
Summary: the conda solver problem in the Python build envs was due to the outdated numpy pins

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/3c18e38a34a584735125d60b633cad7828dfba87/recipe/conda_build_config.yaml#L11-L15

We could have fixed this weeks ago in https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/74

johnkerl:
Re-opening to implement @jdblischak advice

jdblischak (approved):

PR #78, opened 2024-01-30
tiledbsoma 1.7.0 pre-check 2

johnkerl:
Follows on to #77, but incorporating [core 2.19.1](https://github.com/TileDB-Inc/TileDB/releases/tag/2.19.1)

This follows our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases).

PR #79, opened 2024-02-01
tiledbsoma 1.7.0 pre-check #3

johnkerl:
One more check in the vein of #78 before tagging TileDB-SOMA 1.7.0 in GitHub, following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases).

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -110,7 +110,7 @@ outputs:
         - numba >=0.58.1 # [py>37]
         - numba          # [py<=37]
         - attrs >=22.2
-        - somacore >=1.0.6
+        - somacore >=1.0.7
         - scanpy 1.9.*
     test:
       imports:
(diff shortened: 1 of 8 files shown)


johnkerl:
Needs https://github.com/TileDB-Inc/somacore-feedstock/pull/10

recipe/meta.yaml
```diff
@@ -42,7 +42,10 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        - tiledb =2.18.2
+        # Please see
+        # https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/76#discussion_r1457880263
+        # before changing this:
```
jdblischak, line 47: I'm not sure this comment is necessary. We had to do the awkward pinning of `tiledb 2.18.2` because we hadn't updated the numpy pins, which was forcing the solver to install tiledb 2.18.2 in all the Python envs. Now that we've updated the numpy pins, this workaround is no longer needed.

jdblischak (approved):

recipe/meta.yaml
johnkerl: Good call -- thank you!

johnkerl:
Succeeded

PR #83, opened 2024-02-15
Build a linux-64 variant against fmt 9

jdblischak:
Follow up to #82. The cloud Docker images won't solve now that libtiledbsoma was built against fmt 10. This PR adds an additional build variant against fmt 9 for linux-64 only.

The downside is that this is a lot of extra CI. An alternative would be to pin to fmt 9, but then we'll need to remember to remove the pin in the future (as more and more of conda-forge will migrate to fmt 10).

PR #84, opened 2024-02-16
Try #83 with re-render

johnkerl:
`conda smithy rerender --commit auto` on top of #83

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.7.0" %}
-{% set sha256 = "7c712d03bbc511936816130f4e2af1ee6f467f2426c69f118042975a2175fdf0" %}
+{% set version = "1.7.1" %}
+{% set sha256 = "7b92a6aaff4b859f2f402d581cf895c7cf3061fe8291d7d829ebb59da1660ee8" %}
 # This is the SHA256 of
 #   TileDB-SOMA-i.j.k.tar.gz
 # from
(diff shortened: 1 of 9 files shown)


recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.7.0" %}
-{% set sha256 = "7c712d03bbc511936816130f4e2af1ee6f467f2426c69f118042975a2175fdf0" %}
+{% set version = "1.7.1" %}
+{% set sha256 = "7b92a6aaff4b859f2f402d581cf895c7cf3061fe8291d7d829ebb59da1660ee8" %}
```
jdblischak, line 3: I do not understand how this diff was generated. These are already the current values in `main`

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/main/recipe/meta.yaml

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/89a62d76950db7ecfb6909fc443407d2011f36d8/recipe/meta.yaml#L2-L3

So why does Git show that this PR is updating the lines to their current values?

johnkerl: @jdblischak my apologies, manual error. Will resolve tomorrow.

johnkerl: @jdblischak my apologies. I think I hadn't done `git pull` of `main` on the box I used to apply your #83 patch. I'll rebase this PR on top of current `main`.

johnkerl:
[sc-41243]

PR #87, opened 2024-02-20
tiledbsoma 1.7.2 pre-check

johnkerl:


recipe/meta.yaml
```diff
@@ -13,16 +13,16 @@ package:
...
-#  git_depth: -1
-#  # hoping to be 1.7.1 <-- FILL IN HERE
+source:
+  git_url: https://github.com/single-cell-data/TileDB-SOMA.git
+  git_rev: b95a752a335b50a7e08b9dd729748e3c6ccd9b57
+  git_depth: -1
+  # hoping to be 1.7.2 <-- FILL IN HERE
 
 
 build:
```
jdblischak, line 28: Not necessary for this pre-check PR, but reminder that you'll need to reset the build number to 0 when you submit the actual PR for 1.7.2

PR #89, opened 2024-02-26
Upgrade and rerender

johnkerl:
```
```

Resolves recent nightly-build failures

### DIFF
--- a/.scripts/build_steps.sh
+++ b/.scripts/build_steps.sh
@@ -34,9 +34,9 @@ CONDARC
 export CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED=1
 
 mamba install --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
-    pip mamba conda-build boa conda-forge-ci-setup=4
+    pip mamba conda-build conda-forge-ci-setup=4 "conda-build>=24.1"
 mamba update --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
-    pip mamba conda-build boa conda-forge-ci-setup=4
+    pip mamba conda-build conda-forge-ci-setup=4 "conda-build>=24.1"
 
 # set up the condarc
 setup_conda_rc "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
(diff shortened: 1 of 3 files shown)


jdblischak:
The nightly failure is due to the same as I describe in https://github.com/TileDB-Inc/conda-forge-nightly-controller/pull/57

I'll send a PR to resolve it shortly

jdblischak:
Though this was a useful exercise. We have new conda solvers errors that will need to be resolved prior to the next release of tiledbsoma.

```sh
conda_libmamba_solver.exceptions.LibMambaUnsatisfiableError: Encountered problems while solving:
  - package libtiledbsoma-1.7.2-hd763302_0 requires tiledb >=2.19.1,<2.20.0a0, but none of the providers can be installed
├─ libcurl ==8.5.0 hca28451_0 is requested and can be installed;
├─ libgoogle-cloud ==2.21.0 h72bcb37_2 is requested and can be installed;
├─ libtiledbsoma ==1.7.2 hd763302_0 is installable and it requires
│  └─ tiledb >=2.19.1,<2.20.0a0  with the potential options
│     ├─ tiledb 2.19.1 would require
│     │  └─ libgoogle-cloud >=2.12.0,<2.13.0a0 , which conflicts with any installable versions previously reported;
│     ├─ tiledb 2.19.1 would require
│     │  └─ libcurl >=7.88.1,<8.0a0 , which conflicts with any installable versions previously reported;
│     └─ tiledb 2.19.1 would require
│        └─ openssl >=1.1.1w,<1.1.2a , which can be installed;
└─ openssl ==3.2.1 hd590300_0 is not installable because it conflicts with any installable versions previously reported.
```

I think the problem is libgoogle-cloud. We might need to backport an update to libtiledb 2.19.1 to build against a more recent libgoogle-cloud.

jdblischak:
> We might need to backport an update to libtiledb 2.19.1 to build against a more recent libgoogle-cloud.

Unless the next tiledbsoma release builds against libtiledb 2.20, in which case this solver error will hopefully go away (see https://github.com/conda-forge/tiledb-feedstock/pull/239)

johnkerl:
I'll rebase atop #90 once that's merged

jdblischak:
> @jdblischak says https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/90 supersedes this

I fixed the problem with our workflow that triggers the nightly builds. Other ongoing issues include:

* The nightly was failing prior to the conda-smithy problem that I just fixed. This will need to be investigated
* This PR demonstrated that the recipe as it currently exists is no longer solvable (which is beyond frustrating). I have no idea why the previously solvable environment can no longer be solved (and will have to be figured out before any future tiledbsoma releases)

PR #90, opened 2024-02-26
Prevent unnecessary rerendering failures when conda-smithy is updated

jdblischak:
Supersedes #89 

cc @johnkerl 

xref: https://github.com/TileDB-Inc/conda-forge-nightly-controller/pull/57


johnkerl (approved):

jdblischak:
Also note that the nightly-build has been failing since Feb 12th. Though the original error back then was only on osx-64

https://dev.azure.com/TileDB-Inc/CI/_build?definitionId=43&_a=summary&repositoryFilter=39&branchFilter=8732%2C8732%2C8732%2C8732%2C8732%2C8732%2C8732

PR #96, opened 2024-03-07
tiledbsoma 1.8.0

johnkerl:

### DIFF
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -20,6 +20,11 @@ fi
 
 export CXX17FLAGS="-Wno-deprecated-declarations -Wno-deprecated"
 
+# https://conda-forge.org/docs/maintainer/knowledge_base/#newer-c-features-with-old-sdk
+if [[ $target_platform == osx-*  ]]; then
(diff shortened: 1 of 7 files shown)


nguyenv (approved):

recipe/build-r-tiledbsoma.sh
```diff
@@ -20,6 +20,11 @@ fi
 
 export CXX17FLAGS="-Wno-deprecated-declarations -Wno-deprecated"
 
+# https://conda-forge.org/docs/maintainer/knowledge_base/#newer-c-features-with-old-sdk
+if [[ $target_platform == osx-*  ]]; then
+  CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
```
jdblischak, line 26: It's not hurting anything, but we should be able to remove these lines in the future since we already require a newer `MACOSX_DEPLOYMENT_TARGET` and `MACOSX_SDK_VERSION`. My understanding is that you only need this workaround if you want to still target an older macOS SDK version.

johnkerl: Thanks @jdblischak !!

PR #97, opened 2024-03-11
Make `somacore` dependency exact to match `tiledbsoma`

thetorpedodog:
A previous version of tiledb-soma erroneously marked itself as compatible with versions of tilebsoma v1.0.8. This is because the PyPI package (https://github.com/single-cell-data/TileDB-SOMA/blob/1.8.0/apis/python/setup.py#L332) lists itself as only compatible with == a version, but this Conda package listed itself as compatible with >= a version.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -6,7 +6,10 @@
 # from
 #   https://github.com/single-cell-data/TileDB-SOMA/releases/tag/i.j.k
 #
-# See also build number below
+# See also build number below.
+#
+# Also double-check somacore version against TileDB-SOMA's setup.py: this bumps very infrequently
+# but it can.
 
 package:
   name: {{ name }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -29,7 +29,7 @@ source:
 
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


johnkerl (approved):

johnkerl:
> I'll put up a mod

The person who edits `recipe/meta.yaml` is 99% of the time myself. (See also our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases).)  Making an exact `somacore` pin here is an improvement -- it matches TileDB-SOMA's `apis/python/setup.py` -- but it introduces an error-prone step to be addressed. Namely, the change-me spots in this file are well-highlighted. Similarly the spots that bump regularly including core. The `somacore` package, by contrast, is "almost invisible" and it tends to disappear in the `tiledbsoma` shrubbery. A reminder is the right thing to do here.

https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/97/commits/ebc8814bb449530b52e1c5c4040266257b5853ff

johnkerl (approved):
Approved with https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/97/commits/300e393640dce73909a764437691cb24daffc215

jdblischak (approved):

PR #98, opened 2024-03-11
somacore 1.0.9

johnkerl:
For https://github.com/single-cell-data/TileDB-SOMA/issues/1791

In parallel, @ryan-williams will be putting up a PR for TileDB-SOMA's in-package dependency


johnkerl:
Failed CI is the same as #103

jdblischak:
Is this a pre-check PR?

johnkerl:
@jdblischak I should have bumped build number to 1.

That said, given https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/105#issuecomment-2007258035, I'll abandon this PR and move its contents into #105.

PR #99, opened 2024-03-14
[ci skip] Open Issue if nightly Azure build fails

jdblischak:
This PR updates the nightly builds so that they open an Issue if they fail.

I created 3 new [labels](https://github.com/TileDB-Inc/tiledbsoma-feedstock/labels) (which is how my `open-issue` action knows whether to open a new issue or comment on an existing open one)

image

Now an Issue will be opened for any of the following failures:

* The nightly Azure build failed (the most important). These Issues will be labeled `nightly-failure`
* The GitHub Actions workflow to trigger the nightly build failed. These Issues will be labeled `nightly-setup-failure`. These failures are almost always spurious and can usually be fixed with a restart
* The GitHub Actions workflow to check the Azure nightly status failed. These Issues will be labeled `nightly-azure-status-failure`. These errors should be rare

I copy-paste-modified this setup from 

https://github.com/TileDB-Inc/conda-forge-nightly-controller/blob/f78c27c418589b9441b14f63e14889889474f0d6/scripts/check-azure-status.sh
https://github.com/TileDB-Inc/conda-forge-nightly-controller/blob/f78c27c418589b9441b14f63e14889889474f0d6/.github/workflows/check-azure-status.yml
https://github.com/TileDB-Inc/conda-forge-nightly-controller/blob/f78c27c418589b9441b14f63e14889889474f0d6/.github/workflows/tiledb.yml#L60

PR #104, opened 2024-03-18
[ci skip] Do not open Issue for failed Azure nightly build on a fork

jdblischak:
We only need to check the Azure status of the nightly build (and potentially open an Issue to report a failure) on the main feedstock repository. This is creating too much unnecessary noise on my fork.

xref: follow-up to #99


johnkerl (approved):
🔥

PR #110, opened 2024-03-26
tiledbsoma 1.9.0 pre check

johnkerl:
Following our [established procedure](https://www.notion.so/785d165226074071872a8b797133d36c?v=c168da78516e487eac3746d38924fa97&p=d1c7a37761b04cf699a705bb8963dea9&pm=s)

Note: there is still one open TileDB-SOMA PR which is blocking 1.9.0: https://github.com/single-cell-data/TileDB-SOMA/pull/2305

However, I want to flush out any lurking Conda-solver errors in parallel via this pre-check PR.


jdblischak:
Awesome! As I predicted, bumping tiledb 2.21 avoided the conda solver errors we previously observed in the nightly when building against tiledb 2.20 (#108)

johnkerl:
Closing per standard procedure

PR #120, opened 2024-04-01
Trivial typofix in ticket-cutter

johnkerl:


jdblischak (approved):
Good catch!

jdblischak:
To avoid unnecessary CI jobs, I added the prefix `[ci skip]` when I squashed and merged. I also canceled the already running CI jobs for this PR and its internal branch

PR #128, opened 2024-04-09
Add a build-time litmus test for the `arrow` R package

eddelbuettel:
Per a request of @johnkerl a quick test that the `arrow` R package loads and runs.

### DIFF
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -37,4 +37,15 @@ if [[ $target_platform  == osx-arm64 ]]; then
   R_ARGS_EXTRA="$R_ARGS_EXTRA --no-help"
 fi
 
+# Ensure the arrow package loads 'standalone' as a proxy for possible issues
+# arising when it is loaded to check the R package build
+# We call the helper function arrow_info() returning a (classed) list (for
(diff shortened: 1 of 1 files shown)


johnkerl (commented):
🚀 
Thanks @eddelbuettel !!!!!

@jdblischak this LGTM -- since this is a non-recipe PR, do we need to bump the build number in `recipe/meta.yaml` -- ?

ihnorton (approved):

johnkerl:
CI fails are the same as for #127 and are independent of this PR.

johnkerl:
> @jdblischak this LGTM -- since this is a non-recipe PR, do we need to bump the build number in recipe/meta.yaml -- ?

nvm -- I see as precedent https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/120#issuecomment-2029893320

jdblischak:
>  since this is a non-recipe PR, do we need to bump the build number in `recipe/meta.yaml` -- ?

No, no need to bump the build number. It's not affecting the output binary in any way since it is only a diagnostic for help troubleshooting #126. You'd only need to bump the build number if the update to the build script did something like change a compiler flag

johnkerl:
https://github.com/single-cell-data/TileDB-SOMA/issues/2414

[sc-44854]

PR #129, opened 2024-04-09
tiledbsoma-r check #2: `snappy==1.1`

johnkerl:
Another attempt on #126, besides #127

Our working theory :

* A key dependency of `libarrow`, `snappy`, was recently updated
* This broke `libarrow` and any recipe that depended on it
* Folks are already working to clean up the situation on conda-forge, as evidenced in part by the activity at 
  * https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/699
  * https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/700
  * https://github.com/conda-forge/snappy-feedstock/commits/main/ (2024-04-08 commits)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -162,7 +162,7 @@ outputs:
         # https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-macos-sdks
         - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.9") }}  # [osx and x86_64]
         # Temporary https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/126
-        - r-snappy ==1.1
+        - snappy ==1.1
       host:
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
         - r-base
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -161,6 +161,8 @@ outputs:
         - r-rlang                    # [build_platform != target_platform]
         # https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-macos-sdks
         - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.9") }}  # [osx and x86_64]
+        # Temporary https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/126
+        - r-snappy ==1.1
       host:
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
         - r-base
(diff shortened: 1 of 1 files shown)


jdblischak:
This was a useful experiment. As we can see in the [linux-64 build](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=38889&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7), when snappy 1.1 is installed, the R build is fine.

According to the discussion in https://github.com/conda-forge/arrow-cpp-feedstock/issues/1365, the conda-forge metadata has been properly updated, so this workaround should no longer be needed.

For some reason only the [osx-arm64 build](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=38889&view=logs&j=ced5d8de-8227-5f3f-33a1-45cf1592c45a&t=f7bcb349-6dca-53ba-d88d-1d9963e119ba) is still having trouble solving the Python environment due to aws-crt-cpp. Given that all the other environments have passed, my guess is that we need to track down a dependency with a failed osx-arm64 build

```sh
conda_libmamba_solver.conda_build_exceptions.ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-arm64: {MatchSpec("libtiledbsoma==1.9.4=h6ed0caa_0"), MatchSpec("tiledb[version='>=2.21.2,<2.22.0a0']")}
Encountered problems while solving:
  - package libtiledbsoma-1.9.4-h6ed0caa_0 requires tiledb >=2.21.2,<2.22.0a0, but none of the providers can be installed

├─ aws-crt-cpp 0.26.6.* hcff1705_1 is requested and can be installed;
└─ libtiledbsoma 1.9.4.* h6ed0caa_0 is not installable because there are no viable options
   ├─ libtiledbsoma 1.9.4 would require
   │  └─ tiledb >=2.21.2,<2.22.0a0 , which requires
   │     └─ aws-crt-cpp >=0.26.4,<0.26.5.0a0 , which conflicts with any installable versions previously reported;
   └─ libtiledbsoma 1.9.4 conflicts with any installable versions previously reported.
```

jdblischak:
This workaround is no longer needed. The nightly builds passed(https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/126#issuecomment-2047770517) because they installed snappy 1.1 (ie the conda-forge repodata was properly patched). Also, whatever dependency was missing for osx-arm64 when this CI ran, it was available later in the evening for the nightly build.

PR #135, opened 2024-04-17
tiledbsoma 1.10.0

johnkerl:

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.9.5" %}
-{% set sha256 = "f8db75b956ef2c337a08040d9870659c987c9ce5c5b7b771de1e3d4e95c258a8" %}
+{% set version = "1.10.0" %}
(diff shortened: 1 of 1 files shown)


jdblischak (changes requested):
Between [1.9.5 and 1.10.0](https://github.com/single-cell-data/TileDB-SOMA/compare/1.9.5...1.10.0#diff-9fa81ab9f617a2933c22b21b13d62216bd5aedf0a0ee285dc220e4bb4e2f566d), TileDB-SOMA-R added the new dependency nanoarrow. The CI passes because it installs [r-tiledb](https://github.com/conda-forge/r-tiledb-feedstock/blob/05fbe6a580acbc1e957f63d2ab85c08e0f9db58f/recipe/meta.yaml), which already depends on r-nanoarrow. However, we should make this dependency explicit by adding `r-nanoarrow` to the build, host, and run requirements of r-tiledbsoma

johnkerl:
Thanks @jdblischak ! https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/135/commits/58c78c98f3fbb016ad568bcb4edec958df702deb

jdblischak (approved):

PR #138, opened 2024-04-22
Test-drive linux+arm64 [WIP]

johnkerl:
I pushed a branch by @Shelnutt2 and have rebased it ... this is only a WIP draft.

Use cases:

* Anyone doing `docker run` of Linux images on Mac M-series hardware
* Users of AWS Graviton

For: https://github.com/single-cell-data/TileDB-SOMA/issues/2413

Uses: https://github.com/single-cell-data/TileDB-SOMA/pull/2470

PR #141, opened 2024-04-23
Health check from commit hash [do not merge]

johnkerl:
#138 and #139 are probes for a pin update and a new platform, respectively.

They've also got some weird stuff happening that I want to factor away from the probe material per se. Hence #140 and #141.


johnkerl:
@jdblischak addressing via https://github.com/conda-forge/tiledb-feedstock/pull/276 ...

johnkerl:
Success!

PR #161, opened 2024-05-24
Move past tiledbsoma 1.11.2 workaround

johnkerl:
#157 was necessary in order to complete artifact delivery for 1.11.2.

At head of TileDB-SOMA, the patch there is no longer necessary, and in fact, attempts to apply it break nightlies as at #160.


jdblischak (approved):
Removing the patch will cause the linux-64 builds to fail for this PR. I'm fine with that since we want the important signal from our nightly builds.

The alternative would be to update the nightly workflow job to temporarily remove the patch lines from the recipe. But given that we know we are going to remove this patch in the next release, I don't feel that extra effort is required. But that would be the only option to keep the CI green on both `main` and the nightly builds

johnkerl:
@jdblischak please see https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/161/commits/0f5c93dc28a63dc4b625d6987a45a6d86b8a70df -- also I updated the description field of this PR (probably while you were replying in the discussion log).

jdblischak:
> Really up to you. If you can live with red nightly CI until 1.12.0 pre-check, I can ...

This isn't a problem for me in this case because I don't envision it causing us other problems

> This is my preference if you're OK with it. I know how to write `sed` (for 30 years now) but I don't see the one-off PR work as worth the keystroking.

Agreed. I don't think it is worth the effort

PR #171, opened 2024-06-12
tiledbsoma 1.12.0 pre-check

johnkerl:
Following our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases)

Significant dialog is on https://github.com/single-cell-data/TileDB-SOMA/pull/2692

PR #174, opened 2024-06-17
tiledbsoma 1.12

johnkerl:

Leverages all of https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/171 -- the only differences here from that PR are the `source:` bits in `recipe/meta.yaml` to point to https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.12.0

https://github.com/single-cell-data/TileDB-SOMA/issues/2672

### DIFF
--- a/.ci_support/linux_64_fmt10spdlog1.13.yaml
+++ b/.ci_support/linux_64_fmt10spdlog1.13.yaml
@@ -16,12 +16,6 @@ docker_image:
 - quay.io/condaforge/linux-anvil-cos7-x86_64
 fmt:
 - '10'
-numpy:
-- '1.22'
-- '1.23'
-- '1.26'
-- '1.22'
-- '1.22'
 pin_run_as_build:
   python:
     min_pin: x.x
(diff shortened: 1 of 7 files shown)


jdblischak (changes requested):
I don't think we want the `<2` qualifier on numpy. As you noticed when you rerendered, this completely remove the `numpy` field from the `.ci_support/` YAML files. These specify the exact version of numpy that will be installed at build time.

I looked at the build logs of the pre-check PR (#171). After you added the `<2`, this resulted in the latest possible version of numpy 1.x being installed at build time instead of the typically pinned version (a pin used throughout conda-forge, so we don't want to mess with).

The advice I've seen to prevent numpy 2 is to pin in the run requirements. But we are already doing that via `pin_compatible()`.

Could you please remove all the `<2` pins on numpy and rerender? Then we can verify that numpy 2 is never installed

jdblischak (approved):
I recommend doing a quick search of the build log to confirm numpy <2 and pyarrow >=14 . If that sanity check looks good, then I think the PR is good to merge 🚀

PR #176, opened 2024-06-18
Remove lower bound on pyarrow

jdblischak:
Followup to #174. In that PR, we set the lower bound `pyarrow >= 14`. However, we can't do this because TileDB Cloud still has pyarrow 11 installed.

This PR:

* Fixes the conda solver problem by removing the lower bound on pyarrow
* In addition it purposefully builds against pyarrow 11. This is a bit overkill since pyarrow doesn't use run exports. However, I wanted to be confident that it could solve an env with pyarrow 11. We could relax this part in the future
* Completely skip the Python 3.12 builds. We don't need them now, and it is too much trouble at the moment. Once we relax the above, we can probably add them back
* Stopped pyarrrow-hotfix from being installed in the non-cross-compiled build envs
* Migrated the recipe to use [`- {{ stdlib("c") }`](https://conda-forge.org/news/2024/03/24/stdlib-migration/). I've been meaning to do this for awhile

All the builds passed on my [fork](https://dev.azure.com/jdblischak/feedstock-builds/_build/results?buildId=852&view=results)

### DIFF
--- a/.ci_support/linux_64_fmt10spdlog1.13.yaml
+++ b/.ci_support/linux_64_fmt10spdlog1.13.yaml
@@ -2,6 +2,10 @@ c_compiler:
 - gcc
 c_compiler_version:
 - '12'
+c_stdlib:
+- sysroot
+c_stdlib_version:
+- '2.12'
 cdt_name:
 - cos6
 channel_sources:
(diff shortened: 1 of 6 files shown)


johnkerl (approved):
Ideally the `c_stdlib` (the "been meaning to do" bits) would be a separate PR ...

Today, I know we're in a crunch to release ...

PR #189, opened 2024-07-31
tiledbsoma 1.13.0 pre-check

johnkerl:


johnkerl:
CI fai same as #188 which likely has to do with recent conda-forge migration of `fmt` and `spdlog` ...

johnkerl:
Commits 2 & 3 address `fmt` 11 and `spdlog` 1.14 ...

johnkerl:
Most recent CI fail: macos only (x86 and arm64 both):

```
├─ fmt 11.0.1.* h420ef59_0 is requested and can be installed;
└─ libtiledbsoma 1.13.0.* h131f870_0 is not installable because it requires
   └─ fmt >=10.2.1,<11.0a0 , which conflicts with any installable versions previously reported.
```

👀 🤔

johnkerl:
Trying https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/189/commits/217fdb4ac17375e32680a1ffa7a080a349e1aaee ... 🤞

johnkerl:
See also https://conda-forge.org/status/migration/?name=fmt11 ...

johnkerl:
I'm confused. Latest commit https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/189/commits/bcff435da0132606542bab03c32fbe5252127b6e, and priors, fail osx CI with

```
├─ fmt 11.0.1.* h3c5361c_0 is requested and can be installed;
└─ libtiledbsoma 1.13.0.* ha6d8b7b_0 is not installable because it requires
   └─ fmt >=10.2.1,<11.0a0 , which conflicts with any installable versions previously reported.
```

-- yet I can't see where this pin arises.

recipe/conda_build_config.yaml
```diff
@@ -11,17 +11,18 @@ channel_targets:
   - tiledb main
 channel_priority:
   - strict
-# conda-forge has migrated to fmt 10 and spdlog 1.12. We need fmt 9 to install
+# conda-forge has migrated to fmt 11 and spdlog 1.14. We need fmt 9 to install
```
ihnorton, line 14: > We need fmt 9 to install

Is this (still) true? I don't see any pins in the image files for fmt or spdlog, and the feedstock-for-cloud version of the tiledb package doesn't depend on them.

ihnorton:
Seems like the pin comes from [here](https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/189/files#diff-c7cf0e34f41466f367d71791ab69553bf699f1acd371e75d85e356859c68317cR21-R22), but I don't see why.

recipe/conda_build_config.yaml
```diff
@@ -31,8 +31,9 @@ zip_keys:            # [linux]
   - spdlog           # [linux]
   - pyarrow          # [linux]
 # Because of the awkwardness required for the cloud variant above
-# (fmt9+spdlog1.11), we have to migrate the osx-* manually here. Once the
-# conda-forge-pinning-feedstock is updated to spdlog 1.13, we can remove the
-# manual pin below
+# (fmt9+spdlog1.11), we also have to manually pin fmt/spdlog for the osx-*
+# builds
+fmt:                 # [not linux]
+  - 11               # [not linux]
```
jdblischak, line 37: The 2 lines above were the key. Because fmt is also undergoing a migration at the moment, we need to manually pin it for the osx-* builds

jdblischak (approved):
pre-check CI has passed

johnkerl:
Closing per our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases)

PR #200, opened 2024-09-18
tiledbsoma 1.14.1

johnkerl:

See also #2999 and [[sc-53002]](https://app.shortcut.com/tiledb-inc/story/53002/update-projects-to-tiledb-2-26).


recipe/meta.yaml
```diff
@@ -92,28 +89,22 @@ outputs:
...
-        # when the repo's own setup.py won't depend on pyarrow-hotfix anymore
-        - pyarrow-hotfix
       run:
         - {{ pin_compatible('numpy', lower_bound='1.16', upper_bound='1.27') }}
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
         - pandas
         - pyarrow
-        # Needed until https://github.com/single-cell-data/TileDB-SOMA/issues/1926 is resolved,
-        # when the repo's own setup.py won't depend on pyarrow-hotfix anymore
-        - pyarrow-hotfix
```
jdblischak, line 105: I'm not sure exactly how to handle this.

* In the pyarrow tracking Issue, it correctly states that we should add the lower bound `pyarrow >= 14.0.1` when removing pyarrow-hotfix (https://github.com/single-cell-data/TileDB-SOMA/issues/1926#issuecomment-1821208184)
* However, we explicitly build against pyarrow 11 for our TileDB Cloud builds

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/183d8889a30f65475fdddf1a03eb4f261e28b7b7/recipe/conda_build_config.yaml#L25-L27

We obviously can't both pin `>= 14.0.1` in the recipe and build a pyarrow 11 variant.

This is a judgement call. It's my understanding that SOMA doesn't use the problematic code paths anyways. And pyarrow-hotfix is explicitly installed in our cloud Dockerfiles anyways, so it'll still be installed if it's not a SOMA requirement.

https://github.com/TileDB-Inc/TileDB-Cloud-JupyterHub/blob/ad5c32c11efa7dba8787c2cceb5d9887b616be06/notebook/Dockerfile#L171

But it would be possible to install this new SOMA build against pyarrow 11 into a conda env without pyarrow-hotfix installed.

johnkerl: For me I think this suffices -- I've slacked @ihnorton as well to confirm ...

jdblischak: One last thing. I confirmed that the upstream `setup.py` didn't add a lower bound on a pyarrow (which would have prevented building our pyarrow 11 variant for the cloud)

https://github.com/single-cell-data/TileDB-SOMA/blob/1.14.1/apis/python/setup.py#L335

ihnorton: Sounds good to me 👍

johnkerl: Thanks @ihnorton ! :)

jdblischak (approved):

PR #213, opened 2024-10-23
tiledbsoma 1.14.5 pre-check

johnkerl:

[[sc-58115]](https://app.shortcut.com/tiledb-inc/story/58115/reading-obs-previously-updated-errors)


jdblischak (changes requested):
Need to bump somacore to 1.0.21

jdblischak (approved):
Nevermind. Didn't realize that 1.14.x was pinned to 1.0.17. The build passed `pip check`, so all good

https://github.com/single-cell-data/TileDB-SOMA/blob/1.14.4/apis/python/setup.py#L338

johnkerl:
Yup @jdblischak just a single-PR mod :|

PR #214, opened 2024-10-23
tiledbsoma 1.14.5

johnkerl:

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.14.2" %}
-{% set sha256 = "b459976562fdea0c5a2ae03843b44e82e0a1d1f0721f3452d19ba49ad8d62231" %}
+{% set version = "1.14.5" %}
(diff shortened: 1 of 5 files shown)


recipe/meta.yaml
```diff
@@ -23,9 +23,9 @@ source:
...
 #source:
 #  git_url: https://github.com/single-cell-data/TileDB-SOMA.git
-#  git_rev: 2d5d705b020f2996a9266fe9268732f7c440cc37
+#  git_rev: afca166d73cf97a0daf5e73ba6ce976064de0018
 #  git_depth: -1
-#  # hoping to be 1.14.2 <-- FILL IN HERE
+#  # hoping to be 1.14.5 <-- FILL IN HERE
 
 build:
   number: 1
```
jdblischak, line 31:
```suggestion
  number: 0
```

Reset build number to zero for recipe version bump

jdblischak (changes requested):

recipe/meta.yaml
johnkerl: gurk

thanks @jdblischak !

jdblischak (approved):

Issue #212, opened 2024-10-21
Update anndata requirement to `anndata>=0.10.1` for next release


jdblischak:
For the next release, we need to update the anndata requirement to `anndata>=0.10.1`. I'm creating an Issue since otherwise I'm bound to forget.

xref: https://github.com/single-cell-data/TileDB-SOMA/issues/1785, https://github.com/single-cell-data/TileDB-SOMA/pull/3141

While doing that we can also drop the `py<=37` logic since conda-forge dropped py37 support a while ago.

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/21b27689499a1fd8686c16f85ac4431b67dfaae7/recipe/meta.yaml#L102-L103

johnkerl:
Thanks @jdblischak !

PR #218, opened 2024-10-30
tiledbsoma 1.15.0rc3

johnkerl:


recipe/meta.yaml
```diff
@@ -99,15 +99,13 @@ outputs:
         - pyarrow
         - python
         - scipy
-        - anndata       # [py>37]
-        - anndata <0.9  # [py<=37]
+        - anndata
```
jdblischak, line 102:
```suggestion
        - anndata >=0.10.1
```

https://github.com/single-cell-data/TileDB-SOMA/blob/1.15.0rc3/apis/python/setup.py#L330

xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/212

jdblischak (changes requested):

jdblischak (approved):

PR #220, opened 2024-10-31
Manually migrate to numpy 2

jdblischak:
I missed yesterday (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/218) that [1.15.0rc3](https://github.com/single-cell-data/TileDB-SOMA/releases/tag/1.15.0rc3) added support for numpy 2. Since this feedstock isn't part of conda-forge, we need to manually migrate tiledbsoma-feedstock to numpy 2, just like we are doing for tiledb-vcf-feedstock (https://github.com/TileDB-Inc/tiledb-vcf-feedstock/pull/141#pullrequestreview-2408199914)


jdblischak:
Well I'm stuck. We get a conda solver error trying to use numpy2 with our cloud variant (built with outdated pins).

Locally I tried various combinations of defining `python`, `numpy`, and `python_impl` in `conda_build_config.yaml`, but nothing worked. At this point I don't think we can migrate this feedstock to numpy 2 until we drop the cloud variant (at least not without implementing a hacky workaround that would require manual intervention after each conda-smithy rerendering).

jdblischak:
@ihnorton confirmed that we will drop the py39 environment when we transition the cloud to py312+numpy2. Thus the goal of this PR is to test that the feedstock transition to numpy2 will go smoothly in the future when the py39 cloud variant is dropped

jdblischak:
Hmmm...it appears that the [fmt 11](https://conda-forge.org/status/migration/?name=fmt11) and [spdlog 1.14](https://conda-forge.org/status/migration/?name=spdlog114) migrations are stalled. I'll have to manually apply these to the feedstock since they haven't made it to conda-forge-pinning yet

jdblischak:
Also cross-referencing upstream support for numpy 2: https://github.com/single-cell-data/TileDB-SOMA/issues/2536, https://github.com/single-cell-data/TileDB-SOMA/pull/2934

jdblischak:
Went with the 2-branch approach. Superseded by #227 and #228

PR #226, opened 2024-11-18
Produce packages for python 3.12

dudoslav:
In order to solve environments with python 3.12 we require tiledbsoma-py to support python 3.12 as well.

I removed support for pyarrow 11, not sure if that is an issue.

PR #227, opened 2024-11-18
Build py39 cloud variant in a separate branch

jdblischak:
We are transitioning our cloud environment from py39 to py312. It is proving difficult (if not impossible) to configure the feedstock for both the past (py39 with old pins like pyarrow 11) and the future (py312 and numpy2). In order to test the new py312 environment while also maintaining the existing py39 environment, I have created a new branch [py39cloud](https://github.com/TileDB-Inc/tiledbsoma-feedstock/tree/py39cloud).

cc: @dudoslav, @ihnorton 
xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/220, https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/226

This PR configures the feedstock to only build a single linux-64 python 3.9 variant with the old pins for our existing cloud environment. The only real change I made was to change the label for the cloud variant from "main" to "for-cloud". In retrospect, we should have always been doing this, since the cloud variants of libtiledbsoma/tiledbsoma-py/r-tiledbsoma are only valid when installed with the "for-cloud" variant of libtiledb. I confirmed that both the notebook and UDF Docker images specify the channel ` tiledb/label/for-cloud` when installing the SOMA conda binaries, so this won't affect the next cloud deployment.


johnkerl (approved):
LGTM!

jdblischak:
Confirmed that the new branch is able to upload compatible binaries:

```sh
mamba create --dry-run -n test \
  -c tiledb/label/for-cloud -c conda-forge -c tiledb \
  fmt=9 spdlog=1.11 pyarrow=11 python=3.9 numpy=1.22 tiledbsoma-py
## + tiledbsoma-py           1.15.0rc3  py39h6b0a75e_0          tiledb/label/for-cloud      646kB
```

PR #231, opened 2024-11-20
Build r-tiledbsoma with C++ 20

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SOMA/issues/3154

### DIFF
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -43,8 +43,6 @@ spdlog:
 - '1.14'
 target_platform:
 - linux-64
-tiledb:
-- '2.26'
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
--- a/.ci_support/osx_64_.yaml
+++ b/.ci_support/osx_64_.yaml
@@ -45,8 +45,6 @@ spdlog:
 - '1.14'
 target_platform:
 - osx-64
-tiledb:
-- '2.26'
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
--- a/.ci_support/osx_arm64_.yaml
+++ b/.ci_support/osx_arm64_.yaml
@@ -45,8 +45,6 @@ spdlog:
 - '1.14'
 target_platform:
 - osx-arm64
-tiledb:
-- '2.26'
 zip_keys:
 - - c_compiler_version
   - cxx_compiler_version
--- a/build-locally.py
+++ b/build-locally.py
@@ -26,6 +26,13 @@ def setup_environment(ns):
             os.path.dirname(__file__), "miniforge3"
         )
 
+    # The default cache location might not be writable using docker on macOS.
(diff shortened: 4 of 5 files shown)


jdblischak:
For some reason the osx-64 build isn't using C++20. The [errors](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=41247&view=logs&j=e0208569-136d-54da-4ec5-14d0e8771cbc&t=25992e30-3e2a-5788-2b3e-e373b0534bf7) are the same as in the nightly build (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/230#issuecomment-2488876179)

jdblischak (changes requested):
@johnkerl could you please rebase your PR onto the latest from main to include #228?

jdblischak:
Ah, looking earlier in the build log, I see we have a nice diagnostic signal. R explicitly tells us it is using C++17 on osx-64. 

```sh
* installing *source* package ‘tiledbsoma’ ...
** using staged installation
** updated src/Makevars for system library via pkg-config
** libs
using C++ compiler: ‘clang version 18.1.8’
using C++17
using SDK: ‘NA’‘NA’‘NA’‘’
```

Bizarrely though, the osx-arm64 build also reports `using C++17`, so unclear why it is able to compile successfully.

johnkerl:
Commit 3:

```
```

jdblischak:
I think it is going to be tough to build r-tiledbsoma with C++20 without the upstream changes in https://github.com/single-cell-data/TileDB-SOMA/pull/3331 that updated `configure` and `src/Makevars.in`.

If we want to fix the nightly build, I think we'll need to 1) fix the build on the branch [nightly-build](https://github.com/TileDB-Inc/tiledbsoma-feedstock/tree/nightly-build), and then 2) modify [`nightly/update-recipe.sh`](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/main/.github/scripts/nightly/update-recipe.sh) to incorporate these changes.

johnkerl:
@jdblischak I was confused

https://github.com/single-cell-data/TileDB-SOMA/pull/3331 is merged to `main` of TileDB-SOMA & I was confusing this PR with the centralized-nightlies which should be testing the day's `main`.

Let me take a look at `nightly-build` and `nightly/update-recipe.sh` which, embarrassingly, I'm not at all familiar with ...

jdblischak:
> I was confusing this PR with the centralized-nightlies which should be testing the day's `main`.

Good news! The TileDB-SOMA-R build in centralized-tiledb-nightlies is [using C++20](https://github.com/TileDB-Inc/centralized-tiledb-nightlies/actions/runs/11936899558/job/33273058715#step:10:14) and building without error. Thus if we can get the nightly feedstock build to also use C++20, it should _just work_.

johnkerl:
Closed in favor of #246

PR #232, opened 2024-11-22
Pre-check for PR 3361

johnkerl:
Context: https://github.com/single-cell-data/TileDB-SOMA/pull/3361#issuecomment-2492055089 for


johnkerl:
This will need to include #231 at minimum -- I'll restack it now.

johnkerl:
... restacked and force-pushed just now.

PR #235, opened 2024-11-26
Test upstream branch release-1.15

jdblischak:
This demonstrates that the upstream branch release-1.15, which uses C++17, can be compiled in the conda feedstock.

This is only for documentation purposes and should not be merged.

Also note that we will still need to create a pre-check PR once tiledb 2.27 is available as a conda binary.

xref: https://github.com/single-cell-data/TileDB-SOMA/pull/3389, https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/234


jdblischak:
CI passed!

PR #237, opened 2024-12-10
tiledbsoma 1.15.0 pre-check

johnkerl:

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -94,7 +94,7 @@ outputs:
         - pyarrow
         - python
         # https://github.com/single-cell-data/TileDB-SOMA/issues/3444
-        - scipy <0.15.0
+        - scipy
         - anndata >=0.10.1
         - typing-extensions >=4.1
         - numba >=0.58.1
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,7 +24,7 @@ package:
 source:
   git_url: https://github.com/single-cell-data/TileDB-SOMA.git
   # release-1.15 branch 2024-12-16
-  git_rev: 40c5b0ab6bad10926c579b305a426478a165bd5c
+  git_rev: e5d19d05b659cd04f43e5b4a227ace8b5e676815
   git_depth: -1
   # hoping to be 1.15.0 <-- FILL IN HERE
 
@@ -93,20 +93,21 @@ outputs:
         - pandas
         - pyarrow
         - python
-        - scipy
+        # https://github.com/single-cell-data/TileDB-SOMA/issues/3444
+        - scipy <0.15
         - anndata >=0.10.1
-        - tiledb-py >=0.33.0,<0.34.0
         - typing-extensions >=4.1
         - numba >=0.58.1
         - attrs >=22.2
         # Keep this in sync with TileDB-SOMA's somacore version requirement.
-        - somacore ==1.0.25
+        - somacore ==1.0.26
         - scanpy >=1.9.2
     test:
       imports:
         - tiledbsoma
       requires:
         - pip
+        - tiledb-py >=0.33.0,<0.34.0
       commands:
         - python -c 'import tiledbsoma; print(tiledbsoma.pytiledbsoma.version())'
         # See also:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -94,7 +94,7 @@ outputs:
         - pyarrow
         - python
         # https://github.com/single-cell-data/TileDB-SOMA/issues/3444
-        - scipy
+        - scipy <1.15.0
         - anndata >=0.10.1
         - typing-extensions >=4.1
         - numba >=0.58.1
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -94,7 +94,7 @@ outputs:
         - pyarrow
         - python
         # https://github.com/single-cell-data/TileDB-SOMA/issues/3444
-        - scipy <0.15
+        - scipy <0.15.0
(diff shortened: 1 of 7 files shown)


recipe/meta.yaml
```diff
@@ -45,10 +46,6 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        # Workaround for conda solver problems for osx-64. Remove when migrating
-        # to 2.27
-        - tiledb 2.26.1  # [osx and x86]
-        - tiledb >=2.26.0,<2.27  # [not (osx and x86)]
```
jdblischak, line 51: This needs to be `- tiledb >=2.27.0,<2.28` with no selectors

jdblischak (changes requested):
Adding back `tiledb` to the host requirements of libtiledbsoma should fix some of the compilation errors

recipe/meta.yaml
```diff
@@ -45,10 +46,7 @@ outputs:
...
+        - tiledb >=2.27.0,<2.28  # [not (osx and x86)]
```
jdblischak, line 49:
```suggestion
        - tiledb >=2.27.0,<2.28
```

No selectors. Those were only for the osx-64 workaround

recipe/meta.yaml
```diff
@@ -97,21 +95,18 @@ outputs:
         - python
         - scipy
```
jdblischak, line 96:
```suggestion
        - scipy <1.15.0
```

Not currently a problem, but this will avoid future problems just like upstream (xref: https://github.com/single-cell-data/TileDB-SOMA/pull/3445)

recipe/meta.yaml
```diff
@@ -97,21 +95,18 @@ outputs:
...
         - anndata >=0.10.1
-        - tiledb-py >=0.32.0,<0.33.0
+        - tiledb-py >=0.33.0,<0.34.0
```
jdblischak, line 98: TileDB-Py is an optional dependency now, right? I recommend moving this requirement to the test section. That way we can still confirm it can be co-installed with the correct version of tiledb-py, but allow users the possibility to install a minimal tiledbsoma-py conda env.

jdblischak (changes requested):
Mostly looking good now. Just a few minor suggestions to improve the TileDB-Py requirements

recipe/meta.yaml
johnkerl: Thanks @jdblischak !

Let me 'get this green' for pre-check & unblock some things in the [dependency chain](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) tonight & I'll definitely do that moving forward

recipe/meta.yaml
johnkerl: Thanks @jdblischak ! I'll apply this mod, too, for the 'real' feedstock PR coming up after this pre-check passes (hopefully tonight!)

johnkerl:
`osx_64` passed on retry

johnkerl:
> Let me 'get this green' for pre-check & unblock some things in the [dependency chain](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases) tonight & I'll definitely do that moving forward

Derp, I needed to do this too
https://github.com/single-cell-data/TileDB-SOMA/pull/3460
-- so I'll re-run this pre-check with (a) that PR once backported to `release-1.15` and merged; (b) `somacore==1.0.26` here; (c) `scipy<0.15` here; (d) tiledb-py moved to `tests` as long as I can figure it out

johnkerl:
@jdblischak https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/237/commits/7cb1d9f8f410574e42612b5749e2b9dd456bdfd1 has the aformentioned four changes

johnkerl:
This commit
https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/237/commits/721f2fce128bd4136d75f47cac210de8c5d54830
is because 45 minutes (!) into the build on the previous commit I got a failure with

```
2024-12-17T05:50:07.1157920Z Encountered problems while solving:
2024-12-17T05:50:07.1158670Z   - nothing provides scipy <0.15.0 needed by tiledbsoma-py-1.15.0-py310h06877d3_0
2024-12-17T05:50:07.1159270Z   - unsupported request
```

which I cannot with.

jdblischak:
> - nothing provides scipy <0.15.0 needed by tiledbsoma-py-1.15.0-py310h06877d3_0

This is bizarre. Clearly scipy <0.15.0 exists. I can investigate later today, but I agree let's not block the release for this

> is because 45 minutes (!) into the build on the previous commit I got a failure with

This delay is because scipy is only a runtime requirement. Thus this didn't surface until the test section when the runtime dependencies are installed.

jdblischak:
> - nothing provides scipy <0.15.0 needed by tiledbsoma-py-1.15.0-py310h06877d3_0

Figured it out. It's scipy 1.15.0, not 0.15.0

recipe/meta.yaml
```diff
@@ -95,23 +93,21 @@ outputs:
         - pandas
         - pyarrow
         - python
+        # https://github.com/single-cell-data/TileDB-SOMA/issues/3444
         - scipy
```
jdblischak, line 97:
```suggestion
        - scipy <1.15.0
```

johnkerl: @jdblischak thanks for catcthing the typo.

I still fear it won't let me pin under something that doesn't exist _yet_.

I'll spend another 45-minute run trying that ...

jdblischak (commented):
Pre-check looks good! 🚀

PR #238, opened 2024-12-17
tiledbsoma 1.15.0

johnkerl:

[[sc-59686]](https://app.shortcut.com/tiledb-inc/story/59686/update-projects-to-tiledb-2-27)


jdblischak (approved):
🚀

PR #246, opened 2025-01-09
C++20 R build

XanthosXanthopoulos:
After switching to C++20 the R build failed with an cryptic memory error message during build (see #230). The culprit seems to be `-fvisibility-inlines-hidden`, where the build succeeds once the flag is removed. With C++17 there is no issue with that flag and I haven't succeeded in a local repro of the issue outside of a docker container.

### DIFF
--- a/recipe/build-libtiledbsoma.sh
+++ b/recipe/build-libtiledbsoma.sh
@@ -3,7 +3,7 @@
 set -exo pipefail
 
 # Clear default compiler flags
-export CXXFLAGS=""
+export CXXFLAGS=${CXXFLAGS//"-fvisibility-inlines-hidden"/}
 
 mkdir libtiledbsoma-build && cd libtiledbsoma-build
 
--- a/recipe/build-libtiledbsoma.sh
+++ b/recipe/build-libtiledbsoma.sh
@@ -2,6 +2,9 @@
 
 set -exo pipefail
 
+# Clear default compiler flags
+export CXXFLAGS=""
+
 mkdir libtiledbsoma-build && cd libtiledbsoma-build
 
 cmake \
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -5,7 +5,7 @@ set -ex
 cd apis/r
 
 # Clear default compiler flags
-export CXXFLAGS=""
+export CXXFLAGS=${CXXFLAGS//"-fvisibility-inlines-hidden"/}
 
 export DISABLE_AUTOBREW=1
 
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -10,7 +10,7 @@ export CXXFLAGS=""
 export DISABLE_AUTOBREW=1
 
 # https://github.com/conda-forge/r-tiledb-feedstock/commit/29cb6816636e7b5b58545e1407a8f0c29ff9dc39
-if [[ $target_platform  == osx-64 ]]; then
+if [[ $target_platform == osx-* ]]; then
   export NN_CXX_ORIG=$CXX
   export NN_CC_ORIG=$CC
   export CXX=$RECIPE_DIR/cxx_wrap.sh
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -4,6 +4,9 @@ set -ex
 
 cd apis/r
 
+# Clear default compiler flags
+export CXXFLAGS=""
+
 export DISABLE_AUTOBREW=1
 
 # https://github.com/conda-forge/r-tiledb-feedstock/commit/29cb6816636e7b5b58545e1407a8f0c29ff9dc39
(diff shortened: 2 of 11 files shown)


recipe/build-r-tiledbsoma.sh
```diff
@@ -4,21 +4,29 @@ set -ex
...
+if [[ $target_platform == osx-* ]]; then
   export NN_CXX_ORIG=$CXX
   export NN_CC_ORIG=$CC
   export CXX=$RECIPE_DIR/cxx_wrap.sh
   export CC=$RECIPE_DIR/cc_wrap.sh
-  mkdir -p ~/.R
-  echo CC=$RECIPE_DIR/cc_wrap.sh > ~/.R/Makevars
-  echo CXX=$RECIPE_DIR/cxx_wrap.sh >> ~/.R/Makevars
-  echo CXX17=$RECIPE_DIR/cxx_wrap.sh >> ~/.R/Makevars
+  export CXX20=$RECIPE_DIR/cxx_wrap.sh
```
jdblischak, line 18: Do you need to define `CXX20` here when it is overwritten below?

recipe/cc_wrap.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 
-args="${@##-mmacosx-version-min=10.9*}"
-$NN_CC_ORIG $args -mmacosx-version-min=11.0
+args="${@##-mmacosx-version-min=13.3*}"
```
jdblischak, line 3: I don't think the version in these 2 lines shouldn't match. My understanding of the purpose of the wrapper script is to remove the original value of `-mmacosx-version-min` and replace it with a new one. If the script is replacing the exact same string, we can probably remove the wrapper script all-together.

recipe/conda_build_config.yaml
```diff
@@ -1,7 +1,7 @@
 # https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-macos-sdks
 # https://conda-forge.org/news/2024/03/24/stdlib-migration/
-MACOSX_SDK_VERSION:  # [osx and x86_64]
-  - 11.0            # [osx and x86_64]
+MACOSX_SDK_VERSION:            # [osx and x86_64]
+  - 13.3                       # [osx and x86_64]
 c_stdlib_version:              # [osx and x86_64]
   - 11.0                       # [osx and x86_64]
```
jdblischak, line 6: I was concerned that these didn't match, but I think this is correct. From the [conda-forge docs](https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-macos-sdks), we don't need to bump `c_stdlib_version` as long as we're not getting additional errors.

jdblischak (commented):
This is awesome @XanthosXanthopoulos! It's amazing to finally see some green CI for the C++20 builds

johnkerl:
@XanthosXanthopoulos my commits 7,8,9 are as noted in Slack

johnkerl (approved):

recipe/build-libtiledbsoma.sh
```diff
@@ -2,6 +2,9 @@
 
 set -exo pipefail
 
+# Clear default compiler flags
+export CXXFLAGS=${CXXFLAGS//"-fvisibility-inlines-hidden"/}
```
ihnorton, line 6: Where is the flag coming from?

XanthosXanthopoulos: This comes from `activate-gxx_linux-64.sh`  https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=41677&view=logs&j=7b6f2c87-f3a7-5133-8d84-7c03a75d9dfc&t=9eb77fd2-8ddd-5444-8fc0-71cb28dcb736&l=1258

recipe/build-r-tiledbsoma.sh
jdblischak: I confirmed that this instance of `CXX20` is not needed

recipe/cc_wrap.sh
jdblischak: I confirmed that the line `args="${@##-mmacosx-version-min=13.3*}"` doesn't do anything. That's why setting it to the same value that is being added makes no difference. I confirmed that the file can be simplified to:

```sh
#!/bin/sh

$NN_CC_ORIG "$@" -mmacosx-version-min=13.3
```

jdblischak (approved):
I have applied this C++20 patch to the nightly setup workflow in #253

jdblischak:
This PR has served its purpose. When we release a version of SOMA that supports C++20, we'll need to apply these changes to that update PR. I want @XanthosXanthopoulos to get credit for this important contribution, so the following steps should preserve the author information:

```sh
# Apply patch to support C++20
patch -p1 < .github/scripts/nightly/cpp20.patch
# Add and commit
git commit -a -m "Support C++20" --author="XanthosXanthopoulos <xanxanthopoulos@gmail.com>"
# Remove my code to update the nightly builds
git revert 9462dca5141914c05d94624f3b98546223ec5a61
```

PR #250, opened 2025-01-10
tiledbsoma 1.15.3

johnkerl:

[sc-61669](https://app.shortcut.com/tiledb-inc/story/61669/tiledbsoma-1-15-3)

### DIFF
--- a/.ci_support/migrations/spdlog114.yaml
+++ b/.ci_support/migrations/spdlog114.yaml
@@ -5,4 +5,4 @@ __migrator:
   migration_number: 1
 migrator_ts: 1722204008.301365
 spdlog:
-- '1.15'
+- '1.14'
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -40,7 +40,7 @@ python:
 r_base:
 - '4.3'
 spdlog:
-- '1.14'
+- '1.15'
 target_platform:
 - linux-64
 zip_keys:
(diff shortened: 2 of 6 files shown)


.ci_support/migrations/spdlog114.yaml
```diff
@@ -5,4 +5,4 @@ __migrator:
   migration_number: 1
 migrator_ts: 1722204008.301365
 spdlog:
-- '1.14'
+- '1.15'
```
jdblischak, line 8: I find it confusing that we've edited the 1.14 migration file to use 1.15. This migration file is safely superseded by `spdlog115.yaml`. 

Though I agree it is a bit confusing having multiple migration files in the same feedstock, so I'm fine with either 1) deleting `spdlog114.yaml` entirely, or 2) restoring it to its original contents

jdblischak (changes requested):
Just a nit to keep the feedstock infrastructure organized. Otherwise this looks ready

.ci_support/migrations/spdlog114.yaml
johnkerl: Oops, my mistake -- thank you @jdblischak for the code review

johnkerl: @jdblischak https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/250/commits/b8ca57986776d74ad2ff33becbf1efde94ab8991

jdblischak (approved):
🚀

PR #271, opened 2025-02-04
tiledbsoma 1.15.7 pre-check, `py39cloud` branch

johnkerl:

#3662

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "tiledbsoma" %}
-{% set version = "1.15.6" %}
-{% set sha256 = "78ad7959cf2114c8e278f310a5add7ac8b04d97a6141284703755005691fe12d" %}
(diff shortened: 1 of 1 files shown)


jdblischak (approved):
This was built against pyarrow 11, but then tested against pyarrow 19. If we want to test against a specific version of pyarrow (eg to match the version in TileDB Cloud), we could add this to the test requirements

johnkerl:
Thanks @jdblischak ! It's a subset of tests that fail, and I've already tested interactively with `pip install pyarrow==11.0` on a system I control.

That said, to make use of your offer, we'd need to set the test version for `pyarrow`, _and_ ensure we're running the unit-test cases from https://github.com/single-cell-data/TileDB-SOMA/pull/3669 / https://github.com/single-cell-data/TileDB-SOMA/pull/3611 as part of this repo's test coverage.

Issue #279, opened 2025-02-19
Support R 4.4 builds on osx-64 and osx-arm64


jdblischak:
## Background
 
* We recently learned in the nightlies that the r-tiledbsoma builds fail with R 4.4 on osx-64 and osx-arm64 (linux-64 passes) (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/277#issuecomment-2666360359)
* The short-term solution was to prevent the R 4.4 build for now (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/278)
* @johnkerl updated the upstream CI to confirm that TileDB-SOMA-R can be built with R 4.4 on macOS (https://github.com/single-cell-data/TileDB-SOMA/pull/3706)
* This means the problem is specific to conda/conda-forge

## Diagnosis

I figured out the difference between the R 4.3 and 4.4 builds:

* The conda-forge R binary produced by r-base-feedstock switched from using the extension `.dylib` to `.so` between R 4.3 and 4.4 (https://github.com/conda-forge/r-base-feedstock/pull/312)
* Thus we had to apply a [patch](https://github.com/conda-forge/r-tiledb-feedstock/commit/ff8cd7cbb4bccdd1b001594c4d13b99d795431a1) to r-tiledb-feedstock so that it would create `lib/tiledb.dylib` when built with R 4.3 and `lib/tiledb.so` when built with R 4.4 (https://github.com/conda-forge/r-tiledb-feedstock/pull/87#issuecomment-2329664936)
* I confirmed this difference in r-tiledb-feedstock behavior by inspecting the [Azure build logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1111862&view=results) for the most recent commit https://github.com/conda-forge/r-tiledb-feedstock/commit/45fc27cacbdafb33d52ea5b240533dd22c8a70dc
* Thus the r-tiledbsoma builds are failing with R 4.4 because they are expecting `libtiledbsoma.dylib` but `libtiledbsoma.so` is produced (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/277#issuecomment-2663441822)

## Solution(s)?

* The osx-64 failure appears trivial to fix. In the test section we run `otool` on `libtiledbsoma.dylib`, which doesn't exist in the R 4.4 build. I just need to update this to differentiate based on the version of R

[...]

PR #278, opened 2025-02-18
Limit to R 4.3 until TileDB-SOMA-R is tested with R 4.4 on macOS

jdblischak:
Closes #277

xref: https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/277#issuecomment-2666360359

PR #280, opened 2025-02-19
Add R 4.4 builds for linux-64 and osx-64

jdblischak:
The R 4.4 r-tiledbsoma build was already passing on linux-64, and the fix for the osx-64 looked straightforward.

Followup to #278. For more details see #279

### DIFF
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -39,6 +39,7 @@ python:
 - 3.9.* *_cpython
 r_base:
 - '4.3'
+- '4.4'
 spdlog:
 - '1.15'
 target_platform:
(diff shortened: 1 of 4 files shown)


jdblischak:
This PR successfully built r-tiledbsoma for R 4.4 on linux-64 and osx-64. Thus the only remaining problem to troubleshoot is the osx-arm64 build error.

However, I marked it as a draft to not interfere with the imminent 1.16.0 release. We can return to the R 4.4 troubleshooting post-release.

jdblischak:
Rebased onto main https://github.com/TileDB-Inc/tiledbsoma-feedstock/commit/5933255cdb1a8ac09c8761a40465c58a8c87646a to incorporate SOMA 1.16.0 (#288). I purposefully didn't bump the build number because I only want to upload the two new binaries: r-tiledbsoma 1.16.0 with R 4.4 for linux-64 and osx-64

johnkerl (approved):
🚢

PR #275, opened 2025-02-07
Pre-check for 1.16.0 that supports C++20

jdblischak:
Here is the pre-check PR for the future SOMA release that will support C++20. I followed my instructions in https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/246#issuecomment-2604978171 to create the PR.

```sh
git checkout -b jdb/pre-check-cpp20-main

# Apply patch to support C++20
patch -p1 < .github/scripts/nightly/cpp20.patch
## patching file recipe/build-libtiledbsoma.sh
## patching file recipe/build-r-tiledbsoma.sh
## patching file recipe/build-tiledbsoma-py.sh
## patching file recipe/cc_wrap.sh
## patching file recipe/conda_build_config.yaml
## patching file recipe/cxx_wrap.sh
git commit -a -m "Support C++20" --author="XanthosXanthopoulos <xanxanthopoulos@gmail.com>"

# updated recipe to pull from main branch and committed

conda smithy rerender --commit auto

# Remove my code to update the nightly builds
git revert 9462dca5141914c05d94624f3b98546223ec5a61
```

@johnkerl once you have a candidate release ready to pre-check, we just need to update `git_rev` to point to that commit. You can either push directly to my branch (it's from my fork, so you'd have to add my fork as a remote), or you can simply share the SHA here and I can update the PR.

#3720 [[sc-63626]](https://app.shortcut.com/tiledb-inc/story/63626/tiledb-soma-1-16-0)

PR #282, opened 2025-02-21
tiledbsoma-1.16.0 pre-check

johnkerl:

@jdblischak I know I have more to do here following your notes earlier


johnkerl:
Dup of #275

PR #283, opened 2025-03-05
tiledbsoma 1.16.0rc1 pre-check,`main` branch

johnkerl:

https://github.com/single-cell-data/TileDB-SOMA/issues/3720
[[sc-63626]](https://app.shortcut.com/tiledb-inc/story/63626/tiledb-soma-1-16-0)

Sibling of #284. Replays #275.

### DIFF
--- a/.ci_support/osx_64_.yaml
+++ b/.ci_support/osx_64_.yaml
@@ -1,7 +1,7 @@
 MACOSX_DEPLOYMENT_TARGET:
 - '11.0'
 MACOSX_SDK_VERSION:
-- '11.0'
+- '13.3'
 c_compiler:
 - clang
 c_compiler_version:
--- a/.github/scripts/nightly/update-feedstock.sh
+++ b/.github/scripts/nightly/update-feedstock.sh
@@ -16,9 +16,5 @@ sed -i \
   s/"tiledb rc"/"tiledb nightlies"/ \
   recipe/conda_build_config.yaml
 
-# Apply patch from Xanthos to support C++20
-# https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/246
-patch -p1 < .github/scripts/nightly/cpp20.patch
-
 # Print differences
 git --no-pager diff conda-forge.yml recipe/conda_build_config.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@
 
 # Rattler-build's artifacts are in `output` when not specifying anything.
 /output
+# Pixi's configuration
+.pixi
--- a/recipe/build-libtiledbsoma.sh
+++ b/recipe/build-libtiledbsoma.sh
@@ -13,6 +13,7 @@ cmake \
   -DCMAKE_BUILD_TYPE=Release \
   -DTILEDBSOMA_BUILD_CLI=OFF \
   -DTILEDBSOMA_ENABLE_TESTING=OFF \
+  -DSPDLOG_LINK_SHARED=ON \
   ../libtiledbsoma
 
 make -j ${CPU_COUNT}
--- a/recipe/build-libtiledbsoma.sh
+++ b/recipe/build-libtiledbsoma.sh
@@ -2,6 +2,9 @@
 
 set -exo pipefail
 
+# Clear default compiler flags
+export CXXFLAGS=${CXXFLAGS//"-fvisibility-inlines-hidden"/}
+
 mkdir libtiledbsoma-build && cd libtiledbsoma-build
 
 cmake \
--- a/recipe/build-r-tiledbsoma.sh
+++ b/recipe/build-r-tiledbsoma.sh
@@ -4,21 +4,28 @@ set -ex
 
 cd apis/r
 
+# Clear default compiler flags
+export CXXFLAGS=${CXXFLAGS//"-fvisibility-inlines-hidden"/}
+
 export DISABLE_AUTOBREW=1
 
 # https://github.com/conda-forge/r-tiledb-feedstock/commit/29cb6816636e7b5b58545e1407a8f0c29ff9dc39
-if [[ $target_platform  == osx-64 ]]; then
+if [[ $target_platform == osx-* ]]; then
   export NN_CXX_ORIG=$CXX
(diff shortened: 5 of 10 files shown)


johnkerl:
Derp, thanks for the reminders you had been so thorough to set for me, and which I had misplaced. Thank you @jdblischak !

jdblischak (approved):
Not a deal breaker, but it would be nice to add `-DSPDLOG_LINK_SHARED=ON` to `build-libtiledbsoma.sh` to finally fix the dynamic linking to spdlog

johnkerl:
Yup, still iterating with commits -- thanks @jdblischak !

johnkerl:
I'm baffled by the CI fail here:
https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=42007&view=logs&jobId=e0208569-136d-54da-4ec5-14d0e8771cbc&j=e0208569-136d-54da-4ec5-14d0e8771cbc&t=25992e30-3e2a-5788-2b3e-e373b0534bf7
```
conda_libmamba_solver.conda_build_exceptions.ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-64: {MatchSpec("somacore==1.0.28=pyh4471522_0")}
Encountered problems while solving:
  - package somacore-1.0.28-pyh4471522_0 is excluded by strict repo priority
```
-- I'm googling that one now as this is a new one to me

johnkerl:
Baffled by
https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=42007&view=logs&jobId=e0208569-136d-54da-4ec5-14d0e8771cbc&j=e0208569-136d-54da-4ec5-14d0e8771cbc&t=25992e30-3e2a-5788-2b3e-e373b0534bf7
```
conda_libmamba_solver.conda_build_exceptions.ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-64: {MatchSpec("somacore==1.0.28=pyh4471522_0")}
Encountered problems while solving:
  - package somacore-1.0.28-pyh4471522_0 is excluded by strict repo priority
```
-- I have never seen this before.

As a first attempt, I'll rerun the CI just in case the issue involves `somacore` 1.0.28 being too recently published. (Again, I've never seen it fail _in this way_ when a CI here is run too soon, but, it's an easy first thing to try.)

johnkerl:
```
$ conda smithy rerender --commit auto
INFO:conda_smithy.configure_feedstock:Re-rendered with conda-build 25.1.1, conda-smithy 3.46.1, and conda-forge-pinning 2025.03.05.16.19.52
INFO:conda_smithy.configure_feedstock:No changes made. This feedstock is up-to-date.
```

johnkerl:
I think I found the issue: https://github.com/conda-forge/somacore-feedstock now (newly) exists and is still at 1.0.26 while https://github.com/TileDB-Inc/somacore-feedstock is at 1.0.28.

johnkerl:
Regardless we do still want our own channel as higher-pri

johnkerl:
https://github.com/conda-forge/somacore-feedstock/pull/3

johnkerl:
Changing channel priority is bad b/c now a new fail with
```
  - package tiledb-2.27.2-he4dccf3_0 is excluded by strict repo priority
```

So: now that https://github.com/conda-forge/somacore-feedstock exists, I need to add a new release-blocking step to
namely: we must upgrade `conda-forge/somacore-feedstock`, and this is now a new and blocking step in our release process.

johnkerl:
Closing per our [established procedure](https://github.com/single-cell-data/TileDB-SOMA/wiki/Branches-and-releases)

jdblischak:
> So: now that https://github.com/conda-forge/somacore-feedstock exists, I need to add a new release-blocking step to
namely: we must upgrade conda-forge/somacore-feedstock, and this is now a new and blocking step in our release process.

Nice troubleshooting @johnkerl! I hadn't considered this potential problem when somacore and tiledbsoma-py were added upstream to conda-forge. This means that the somacore installed in these feedstock builds is the one from conda-forge.

In light of this:

* Do we need to continue maintaining our own TileDB-Inc/somacore-feedstock? If there is nothing special about it, it seems like an unnecessary step now that we can obtain the conda binary from conda-forge
* Given the complexity of this repo, I'm open to switching the `channel_priority` to "flexible" if we want to be able to install somacore from the tiledb channel

jdblischak:
> Changing channel priority is bad b/c now a new fail with  `- package tiledb-2.27.2-he4dccf3_0 is excluded by strict repo priority`

This puzzled me at first glance, but now I understand it. The `tiledb` channel is for special binaries like "for-cloud" and "nightlies". The most recent standard version we uploaded with the default label of "main" was [2.17.1](https://anaconda.org/tiledb/tiledb/files?channel=main&version=2.17.1) over a year ago.

johnkerl:
> Do we need to continue maintaining our own TileDB-Inc/somacore-feedstock? If there is nothing special about it, it seems like an unnecessary step now that we can obtain the conda binary from conda-forge

Agreed! I've marked this as a follow-up item for myself after we get 1.16.0 out.

> Given the complexity of this repo, I'm open to switching the channel_priority to "flexible" if we want to be able to install somacore from the tiledb channel

I'm open to Conda-expert advice (I am not a Conda expert) but the googling I did do told me strict priority is important for solve times, and I don't want slower solve times.

jdblischak:
> strict priority is important for solve times, and I don't want slower solve times.

Good point! If we're fine with installing somacore from conda-forge, let's keep the strict channel priority.

Personally, the reason strict channel priority is important is because otherwise users that install packages from many different channels end up with "frankenstein" conda environments where their packages come from a mishmash of different places. This leads to hard-to-decipher errors (because packages from different channels won't necessarily be compatible), and also makes it hard to ever reproduce that conda environment. It's much safer to install almost everything from a single channel (eg conda-forge) and then only diverge from this for specific packages you need from other special channels.

johnkerl:
> frankenstein

💯

johnkerl:
Need to test-drive spatial dependencies for [[sc-64100]](https://app.shortcut.com/tiledb-inc/story/64100/)

PR #288, opened 2025-03-10
tiledbsoma 1.16.0, `main` branch

johnkerl:

https://github.com/single-cell-data/TileDB-SOMA/issues/3720 [[sc-63626]](https://app.shortcut.com/tiledb-inc/story/63626/tiledb-soma-1-16-0)

SIbling: #289

### DIFF
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -11,6 +11,6 @@ channel_targets:
 channel_priority:
   - strict
 # Limit to R 4.3 until TileDB-SOMA-R is tested with R 4.4 on macOS
+# Please see https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/279
 r_base:
   - 4.3
-  - 4.4
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -13,3 +13,4 @@ channel_priority:
 # Limit to R 4.3 until TileDB-SOMA-R is tested with R 4.4 on macOS
 r_base:
   - 4.3
+  - 4.4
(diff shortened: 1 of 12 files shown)


recipe/conda_build_config.yaml
```diff
@@ -13,3 +13,4 @@ channel_priority:
 # Limit to R 4.3 until TileDB-SOMA-R is tested with R 4.4 on macOS
 r_base:
   - 4.3
+  - 4.4
```
jdblischak, line 16: We don't want the R 4.4 build on the main branch. We know that osx-64 and osx-arm64 are failing with R 4.4 (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/279). We are building for R 4.4 on the py39cloud branch only, which I think should be sufficient for the immediate needs of our customers

recipe/meta.yaml
```diff
@@ -99,8 +99,14 @@ outputs:
         - numba >=0.58.1
         - attrs >=22.2
         # Keep this in sync with TileDB-SOMA's somacore version requirement.
-        - somacore ==1.0.26
+        - somacore ==1.0.28
         - scanpy >=1.9.2
+        # See TileDB-SOMA's apis/python/requirements_spatial.txt
```
jdblischak, line 104: I thought we were saving the spatial experiment for a later date. spatialdata 0.2.5 still isn't available from conda-forge (I'm actively working on it though)

johnkerl: You're right. I made a mistake & copied too much in. I'll fix this.

recipe/conda_build_config.yaml
johnkerl: Thanks!

recipe/meta.yaml
johnkerl: (These PRs are both in draft)

PR #292, opened 2025-03-18
tiledbsoma 1.16.1, `main` branch

johnkerl:

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -165,7 +165,7 @@ outputs:
         - r-base
         - r-r6
         - r-matrix
-        - r-bit64
+        - r-bit64 >=4.6.0.1          # [build_platform != target_platform]
         - r-rcppint64
         - r-tiledb >=0.31.0,<0.32
         - r-arrow
@@ -183,7 +183,7 @@ outputs:
         - r-base
         - r-r6
         - r-matrix
-        - r-bit64
+        - r-bit64 >=4.6.0.1          # [build_platform != target_platform]
         - r-tiledb >=0.31.0,<0.32
         - r-arrow
         - r-fs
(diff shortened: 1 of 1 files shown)


jdblischak (approved):
This would be a natural opportunity to add the lower bound requirement introduced in 1.16.0: `r-bit64 >=4.6.0.1`. Not a huge deal though if you are in a rush. I can always send a follow-up PR

recipe/meta.yaml
```diff
@@ -148,7 +148,7 @@ outputs:
         - cross-r-base {{ r_base }}  # [build_platform != target_platform]
         - r-rcppspdlog >=0.0.19      # [build_platform != target_platform]
         - r-matrix                   # [build_platform != target_platform]
-        - r-bit64                    # [build_platform != target_platform]
+        - r-bit64 >=4.6.0.1          # [build_platform != target_platform]
```
jdblischak, line 151: Due to the quirks of conda and R, this pin only affects the cross-compiled builds. Unfortunately you have to find-replace 3 times (build/host/run)

recipe/meta.yaml
```diff
@@ -165,7 +165,7 @@ outputs:
         - r-base
         - r-r6
         - r-matrix
-        - r-bit64
+        - r-bit64 >=4.6.0.1          # [build_platform != target_platform]
```
jdblischak, line 168:
```suggestion
        - r-bit64 >=4.6.0.1
```

recipe/meta.yaml
```diff
@@ -183,7 +183,7 @@ outputs:
...
```
jdblischak, line 186:
```suggestion
        - r-bit64 >=4.6.0.1
```

PR #305, opened 2025-04-10
tiledbsoma 1.16.2rc0 pre-check, `main` branch

jp-dark:

### DIFF
--- a/.ci_support/osx_64_.yaml
+++ b/.ci_support/osx_64_.yaml
@@ -1,7 +1,7 @@
 MACOSX_DEPLOYMENT_TARGET:
 - '11.0'
 MACOSX_SDK_VERSION:
-- '13.3'
+- '11.0'
 c_compiler:
 - clang
 c_compiler_version:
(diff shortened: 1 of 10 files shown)


jdblischak (changes requested):
From the [diff to 1.16.1](https://github.com/single-cell-data/TileDB-SOMA/compare/1.16.1...release-1.16), I see that the minimum required version of r-rcppspdlog needs to be bumped to `>=0.0.20`. Strangely this is only enforced for the osx-arm64 build. If this is important, I recommend adding it to the host requirements as well.

Please also include the updates to the build scripts from #299

jdblischak (changes requested):
I apologize for being unclear. You also need to make the changes from #299 to `cc_wrap.sh`, `cxx_wrap.sh`, and `conda_build_config.yaml`. After you commit those changes, then you need to rerender again (`conda smithy rerender --commit auto`) to update the files in `.ci_support/`.

jp-dark:
> After you commit those changes, then you need to rerender again (`conda smithy rerender --commit auto`) to update the files in `.ci_support/`.

Rerendering I get a message that the feedstock is up-to-date. Maybe I missed a change somewhere?

jdblischak (approved):

PR #308, opened 2025-04-28
tiledbsoma 1.16.2, `py39cloud` branch

jp-dark:
Add release 1.16.2 to the conda feedstock


jdblischak:
The tests failed with R 4.4.3. It was a core dump, so we can't see which test failed:

```R
> testthat::test_file('apis/r/tests/testthat.R', stop_on_failure = TRUE)

══ Testing testthat.R ══════════════════════════════════════════════════════════

[ FAIL 0 | WARN 0 | SKIP 0 | PASS 0 ]tiledbsoma:    1.16.1
tiledb-r:      0.31.0
tiledb core:   2.27.2
libtiledbsoma: 2.27.2
R:             R version 4.4.3 (2025-02-28)
OS:            AlmaLinux 9.5 (Teal Serval)
✔ | F W  S  OK | Context

free(): invalid size
/home/conda/feedstock_root/build_artifacts/tiledbsoma_1745873559037/test_tmp/run_test.sh: line 10:  4332 Aborted                 (core dumped) $R -e "testthat::test_file('apis/r/tests/testthat.R', stop_on_failure = TRUE)"
```

jdblischak:
Ideas/thoughts for troubleshooting:

* The tests passed with R 4.4.3 in #307. Thus this suggests the problem is related to the statically linked `libtiledb` used in this branch
* r-tiledb is also failing with R 4.4.3 (passes with 4.4.2) (https://github.com/conda-forge/r-tiledb-feedstock/pull/96, https://github.com/conda-forge/r-tiledb-feedstock/pull/97)
* The cloud env currently only supports R 4.3. Thus short-term we could drop the R 4.4 build. But we do want to migrate to R 4.4 as soon-as-possible, so we'll need to fix this sooner rather than later anyways

recipe/cc_wrap.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-$NN_CC_ORIG "$@" -mmacosx-version-min=13.3
+$NN_CC_ORIG "$@" -mmacosx-version-min=11.0
```
jdblischak, line 3: To reduce the complexity of this branch, you can delete `cc_wrap.sh`, `cxx_wrap.sh`, and the code in `build-r-tiledbsoma.sh` that uses them. Those files are only needed for osx-* builds, and thus they have no effect on this branch to produce a linux-64 conda binary for the cloud env.

jp-dark: Do you mind if I leave them in or do you actively want me to remove them? My thought was to keep the branch as sync with `main` as possible, but I can remove if you have a strong opinion either way.

jdblischak: Your call. I find these diff's distracting because they suggest the changes are relevant to the PR, but they have zero effect on the build. But I see your point about keeping the two branches similar to make it easier to apply changes in parallel to the two branches in the future.

johnkerl:
Editing the `git_rev` in `recipe/meta.yml` and bisecting along `release-1.16` between 1.16.1 and 1.16.2 of TileDB-SOMA isolated the fault (repro using `./build-locally.py` on Ubuntu) to https://github.com/single-cell-data/TileDB-SOMA/pull/3801 which was (coincidentally) the first PR backported to `release-1.16` after 1.16.1.

jdblischak (approved):

PR #325, opened 2025-07-28
TileDB-SOMA 1.17.1 release [WIP]

jp-dark:
Update conda for release 1.17.1


.ci_support/linux_64_.yaml
```diff
@@ -15,15 +15,13 @@ channel_targets:
...
 docker_image:
 - quay.io/condaforge/linux-anvil-x86_64:alma9
 fmt:
 - '11'
 numpy:
-- '2.0'
-- '2.0'
-- '2.0'
+- '2'
 - '2.0'
```
jdblischak, line 25: Heads up @jp-dark that this duplicate numpy pin caused two 1.17.1 binaries to be built and uploaded for every Python version (more explanation in https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/332#issuecomment-3229472338)

Could you please delete the stale migration files that conda-smithy has locally cached on your machine?

```sh
rm -r ~/.cache/conda-smithy/
```

PR #329, opened 2025-08-07
Update anndata to at least version 0.12

jp-dark:
Update anndata to at least version 0.12

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -94,7 +94,7 @@ outputs:
         - pyarrow
         - python
         - scipy
-        - anndata >=0.10.1
+        - anndata >=0.12
         - typing-extensions >=4.1
         - numba >=0.58.1
         - attrs >=22.2


jdblischak:
conda-forge didn't build anndata 0.12 for python 3.9


```sh
conda create --dry-run -n test python=3.9 'anndata>=0.12'
## LibMambaUnsatisfiableError: Encountered problems while solving:
##   - package anndata-0.12.0-pyhd8ed1ab_0 requires python >=3.11, but none of the providers can be installed
## 
## Could not solve for environment specs
## The following packages are incompatible
## ├─ anndata >=0.12 * is installable and it requires
## │  └─ python >=3.11 *, which can be installed;
## └─ python =3.9 * is not installable because it conflicts with any installable versions previously reported.
```

recipe/meta.yaml
```diff
@@ -94,7 +94,7 @@ outputs:
         - pyarrow
         - python
         - scipy
-        - anndata >=0.10.1
+        - anndata >=0.12
```
jdblischak, line 97: anndata dropped support for Python 3.9 upstream while ago in https://github.com/scverse/anndata/commit/dd851287bc996b674f0f35b8f15d33fe2c479b6b (https://github.com/scverse/anndata/pull/1712).

Note that they are only dropping older python versions to follow spec-0 as other scientific python packages are doing (https://github.com/ome/ome-zarr-py/issues/455). In other words, you could likely patch `pyproject.toml` to remove that minimal requirement and build your own custom copy. But that would require forking the conda-forge feedstock to TileDB-Inc and then configuring it to build and upload to the `tiledb` channel

jdblischak (commented):
Reminder that you need to bump the build number to upload a new binary to anaconda.org

PR #332, opened 2025-08-25
Manual opt-in to continue building python 3.9

jdblischak:
xref: https://github.com/conda-forge/tiledb-py-feedstock/pull/270

I didn't rerender because I don't want to introduce any other new changes. This simply keeps building the same 4 Python versions.

Unfortunately it is likely to fail with the same import error as the nightly feedstock builds (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/331)


jdblischak:
Hmmm...for some reason it is building two separate variants for each Python version. I'll need to fix that

```
BUILD START: ['libtiledbsoma-1.17.1-h579f66e_0.conda', 'r-tiledbsoma-1.17.1-r43h04e722a_0.conda', 'r-tiledbsoma-1.17.1-r44h04e722a_0.conda', 'tiledbsoma-py-1.17.1-py39h9613d13_0.conda', 'tiledbsoma-py-1.17.1-py312hae533ea_0.conda', 'tiledbsoma-py-1.17.1-py311hb3f30f9_0.conda', 'tiledbsoma-py-1.17.1-py310hbd70382_0.conda', 'tiledbsoma-py-1.17.1-py39h4e37c57_0.conda', 'tiledbsoma-py-1.17.1-py312head7ac3_0.conda', 'tiledbsoma-py-1.17.1-py311h61cbc2c_0.conda', 'tiledbsoma-py-1.17.1-py310h5e79800_0.conda']
```

jdblischak:
> Hmmm...for some reason it is building two separate variants for each Python version. I'll need to fix that

I found the problem. Somehow in the rererendering step (https://github.com/TileDB-Inc/tiledbsoma-feedstock/commit/365287c152962d4c3f04d3eb055f5945c1eea0d1) when preparing 1.17.1 (https://github.com/TileDB-Inc/tiledbsoma-feedstock/pull/325), a duplicate numpy pin was maintained. I suspect it related to local caching of the numpy2 migration file by conda-smithy (https://github.com/conda-forge/conda-smithy/issues/1833).

https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/677ada89952aa1d17c58ff7ab17cae52773c6d47/.ci_support/linux_64_.yaml#L23-L25

Indeed there were two copies of 1.17.1 uploaded for each Python variant

https://anaconda.org/tiledb/tiledbsoma-py/files?version=1.17.1

And this problem doesn't affect the nightly feedstock build because it rerenders in a clean GitHub Actions run, and thus deletes the stale migration files (https://github.com/TileDB-Inc/tiledbsoma-feedstock/commit/174e08d9a7808dc84cd0497da0c86c9ab7e15fb1)

jdblischak:
I confirmed that the build strings match the existing uploaded binaries, so no new binaries will be uploaded when this PR is merged. This was what I expected since binaries were already built with a numpy pin of `'2'`.

PR #342, opened 2025-10-16
TileDB-SOMA 2.1.0

jp-dark:
TileDB-SOMA 2.1.0 release

PR #343, opened 2025-10-16
Migrate to fmt 12 and spdlog 16

jdblischak:
To match upstream tiledb-feedstock (https://github.com/conda-forge/tiledb-feedstock/pull/465)

xref: #340, #342


jdblischak:
Oops. Only libtiledb 2.29 was migrated to fmt 12. This needs to be done in conjunction with the SOMA 2.1.0 update

PR #352, opened 2025-11-20
Pin libtiledb 2.29.2 for now, due to SOMA-720

ihnorton:
There is a de-facto runtime pin against the build-time version of libtiledb via SOMA's `_verify_expected_tiledb_version` function, which uses a C++ function embedding the libtiledb build-time version triple (from tiledb.h). So this package must enforce an exact pin.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -28,7 +28,7 @@ source:
 #  git_depth: -1
 
 build:
-  number: 0
+  number: 1
   skip: true  # [win or linux32 or py2k]
 # Important: set this back to 0 on a new release
 
@@ -45,7 +45,7 @@ outputs:
         - cmake
         - make  # [not win]
       host:
-        - tiledb >=2.29.0,<2.30
+        - tiledb ==2.29.2
         - spdlog
         - fmt
     about:


jdblischak:
> Is this strictness really required for SOMA?

I've thought more about this. Enforcing versions of libtiledb should not be done at runtime. This should be enforced during installation.

Conda recipes provide "run exports" to ensure that the runtime version of a library is compatible with the version it was built against. Currently the recipe in tiledb-feedstock is set to use a [run exports max pin of `'x.x'`](https://github.com/conda-forge/tiledb-feedstock/blob/6898b5094366f8cb21a6a2bf7b04a3a3c531b657/recipe/meta.yaml#L18). In other words, it allows future patch releases of the same MAJOR.MINOR series to be installed at runtime. If libtiledb is not stable across patch releases, we should update the max pin to `'x.x.x'`. The downsides are that this will require rebuilding all the downstream packages more often, and it also more tightly constrains the conda solver.

https://conda-forge.org/docs/maintainer/pinning_deps/#specifying-run_exports
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

PR #367, opened 2026-01-26
TileDB-SOMA 2.3.0 release

jp-dark:
Update for release 2.3.0.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,7 +24,7 @@ package:
 source:
   git_url: https://github.com/single-cell-data/TileDB-SOMA.git
   # release-2.3.0rc0
-  git_rev: da72c648b3058e8d93d42b345c37193762ccdb81
+  git_rev: aac88138d12a5fb87ed2601500fdab103ba8a8a0
   git_depth: -1
 
 build:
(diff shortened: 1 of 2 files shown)


jdblischak (commented):
The same TileDB-SOMA-R tests are failing on linux-64 as discovered in the nightly feedstock builds (https://github.com/TileDB-Inc/tiledbsoma-feedstock/issues/366)

jdblischak (approved):
🚀 

https://github.com/single-cell-data/TileDB-SOMA/compare/2.2.0...aac88138d12a5fb87ed2601500fdab103ba8a8a0

recipe/meta.yaml
```diff
@@ -89,7 +89,7 @@ outputs:
         - pyarrow
       run:
         - {{ pin_subpackage('libtiledbsoma', exact=True) }}
-        - pandas
+        - pandas <=3.0.0
```
jdblischak, line 92: I just noticed that this is `<=` and not `<` like it is [upstream](https://github.com/single-cell-data/TileDB-SOMA/blob/2.3.0/apis/python/setup.py#L347). Is it ok to install pandas 3 or not? It looks like pandas 2.3.3 was installed, but that is presumably due to some other constraint (which could change at any time).

jp-dark: It should be `<` not `<=`. Good catch.

PR #371, opened 2026-01-28
Fix pandas to be strictly less than 3 [py39cloud]

jp-dark:
Fix a typo in the TileDB-SOMA 2.3.0 release.


ihnorton (approved):

jdblischak (approved):
FYI though that no new binary will be uploaded unless the build number is bumped

